PR #5, opened 2020-02-15
Fixed a few things 

dstarr:
Fixed broken image link
Project names
Added instruction to create 2nd Web App for market provisioning service


dstarr:
Created new solution image to reflect all 5 projects with new project names.

PR #8, opened 2020-02-16
update app settings small fix and install-instructions align

santhoshbomma9:
minor fixes on instructions and appsettings

### DIFF
--- a/docs/Installation-Instructions.md
+++ b/docs/Installation-Instructions.md
@@ -152,7 +152,7 @@ In this section, we will go over the steps to download the latest sources from t
     "ClientId": "<Azure-AD-Application-ID>",
     "ClientSecret": "******",
     "Resource": "62d94f6c-d599-489b-a797-3e10e42fbe22",
-    "FulFillmentAPIBaseURL": "https://marketplaceapi.microsoft.com/",
+    "FulFillmentAPIBaseURL": "https://marketplaceapi.microsoft.com/api/",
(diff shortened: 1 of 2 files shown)


santhoshbomma9:
> I don't think we should be adding that / (slash) at the end of the URL.

Sure, if we aren't adding there, they maybe we should add here when using it?
image

code4clouds:
Typically the URL will not have the / at the end and the code where is reference will have the slash as is {subscriptionBaseURL}/resolve?api .... to indicate is a route within the base url.

code4clouds:
@santhoshbomma9  I also noticed that not using the slash at the end makes your .RESOLVE and .ACTIVE inconsistent.

santhoshbomma9:
> @santhoshbomma9 I also noticed that not using the slash at the end makes your .RESOLVE and .ACTIVE returns (syntax) inconsistent.

And also inconsistent with the appsetting we already have in `Microsoft.Marketplace.SaaS.SDK.CustomerProvisioning` project

code4clouds:
Yeap... @dev-kiranbanda

dev-kiranbanda:
> Yeap... @dev-kiranbanda

Okay, I can remove the / at the end in the config and have "/" prefixed when we build the full URL for the actions like Resolve and like.

If my understanding of the recommendation is correct, I can have the change in my already open PR (if that is okay ).

PR #14, opened 2020-03-10
Updates to configuration and relevant code

dev-kiranbanda:
- Fixed the FulfillmentApiBaseURL in appSettings.json
- Updated the URL helper to prepend "/" while building the URLs for Fulfillment and metering apis


src/Microsoft.Marketplace.SaaS.SDK.Client/Helpers/UrlHelper.cs
```diff
@@ -20,7 +20,7 @@ public static string GetSaaSApiUrl(SaaSApiClientConfiguration clientConfiguratio
         {
             var resourceId = Convert.ToString(resourceGuid);
             string operationId = string.Empty;
-            string subscriptionBaseURL = "saas/subscriptions/";
+            string subscriptionBaseURL = "/saas/subscriptions/";
```
code4clouds: This is also ending with a bar... the bar needs to be carry over to the code that implements the subscriptionBaseURL.

dev-kiranbanda: Sure, however this is internal implementation so should not affect user / configuration

code4clouds: Agree... but its best practice for REST APIs.  Also the variable should be subscriptionBasePath as is not longer a URL.  Not picking the code... just keep them in mind in the future.

dev-kiranbanda: Sure, thank you for the inputs, taken a note of it.

PR #31, opened 2020-04-16
Update API to handle multiple pages of subscriptions.

scseely:
This update handles paging on Subscriptions. One requires paging when the number of subscriptions on a given Commercial Marketplace account exceeds 85 across all offers. 

This issue also identified a problem with URL construction for the all subscriptions path. The default FulFillmentAPIBaseURL in the test was changed to point to the "right" URL.


code4clouds:
@scseely How are you testing this?  Would you share a screenshot?

scseely:
Tested this by running the small set of unit tests. 
Updated the test appsettings.json to point to the Industry Experiences team account which has >150 subscriptions which had been activated.

PR #32, opened 2020-04-16
Fix for warnings

dev-kiranbanda:
Fixed warnings related to render partial.

### DIFF
--- a/src/SaaS.SDK.Client.DataAccess/Context/SaasKitContext.cs
+++ b/src/SaaS.SDK.Client.DataAccess/Context/SaasKitContext.cs
@@ -31,11 +31,7 @@ namespace Microsoft.Marketplace.SaasKit.Client.DataAccess.Context
 
         protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         {
-            if (!optionsBuilder.IsConfigured)
-            {
-#warning To protect potentially sensitive information in your connection string, you should move it out of source code. See http://go.microsoft.com/fwlink/?LinkId=723263 for guidance on storing connection strings.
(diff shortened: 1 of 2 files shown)


src/SaaS.SDK.Client.DataAccess/Context/SaasKitContext.cs
```diff
@@ -31,11 +31,7 @@ public SaasKitContext(DbContextOptions<SaasKitContext> options)
 
         protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         {
-            if (!optionsBuilder.IsConfigured)
-            {
-#warning To protect potentially sensitive information in your connection string, you should move it out of source code. See http://go.microsoft.com/fwlink/?LinkId=723263 for guidance on storing connection strings.
-                optionsBuilder.UseSqlServer("Server=localhost;Initial Catalog=AMPSaaSDB;Persist Security Info=True;User ID=sa;Password=Sa1;");
-            }
+
```
code4clouds, line 34: What/Where are the default SQL values now if this got removed?  

I think this should have stay as its a valid warning, at least till we move the default values to an environment variable or leave them in the .json file as default/sample.

dev-kiranbanda: Sure, but do we want to keep the default values at all like this. The SaasKitContext is initialized at the start up.cs like this:

```csharp

services.AddDbContext<SaasKitContext>(options =>
               options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));

```

If we were to have a default value, I presume, it should be from the json itself, please advise.

PR #46, opened 2020-05-07
fix update subscription status on IsAutomaticProvisioningSupported

santhoshbomma9:
when IsAutomaticProvisioningSupported is true , subscirption status should be Subscribed.
when IsAutomaticProvisioningSupported is false , subscirption status should be PendingActivation and Partner should activate the subscription.
Looks like the logic is a little wrong in the current code. 
Please correct me if my understanding is wrong on the Appconfig setting.

### DIFF
--- a/src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
+++ b/src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
@@ -375,7 +375,7 @@
                         try
                         {
                             this.logger.LogInformation("operation == Activate");
-                            if (Convert.ToBoolean(applicationConfigRepository.GetValuefromApplicationConfig("IsAutomaticProvisioningSupported")))
+                            if (!Convert.ToBoolean(applicationConfigRepository.GetValuefromApplicationConfig("IsAutomaticProvisioningSupported")))
                             {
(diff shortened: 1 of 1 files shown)


src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
```diff
@@ -375,7 +375,7 @@ public IActionResult SubscriptionOperation(Guid subscriptionId, string planId, s
                         try
                         {
                             this.logger.LogInformation("operation == Activate");
-                            if (Convert.ToBoolean(applicationConfigRepository.GetValuefromApplicationConfig("IsAutomaticProvisioningSupported")))
```
dev-kiranbanda, line 378: Yes, you are right. However, this is the not the only place that requires a correction. We are checking this in all the relevant places where this flag value is relevant ( like menu on the subscription detail page on the customer portal ).

santhoshbomma9: Sure @dev-kiranbanda, I checked other places and looks like the logic looks right at other places.

dev-kiranbanda: We are good to go then :)

PR #47, opened 2020-05-08
Quantity fix

santhoshbomma9:
When customer comes to the landing page - removed quantity value from subscription api is being reset to 0. For peruser plans, they always come with a quantity >0 and the remaining it is serialized to zero anyway.

### DIFF
--- a/src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
+++ b/src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
@@ -139,7 +139,6 @@
                             this.subscriptionService.AddPlanDetailsForSubscription(subscriptionPlanDetail);
                             // GetSubscriptionBy SubscriptionId
                             var subscriptionData = this.apiClient.GetSubscriptionByIdAsync(newSubscription.SubscriptionId).ConfigureAwait(false).GetAwaiter().GetResult();
-                            subscriptionData.Quantity = 0;
                             var subscribeId = this.subscriptionService.AddUpdatePartnerSubscriptions(subscriptionData);
                             if (subscribeId > 0 && subscriptionData.SaasSubscriptionStatus == SubscriptionStatusEnum.PendingFulfillmentStart)
                             {
--- a/src/SaaS.SDK.CustomerProvisioning/Views/Home/Index.cshtml
+++ b/src/SaaS.SDK.CustomerProvisioning/Views/Home/Index.cshtml
@@ -59,7 +59,7 @@
                         <dd class="col-sm-9 p-2 p10">
                             @Html.DisplayFor(model => model.PlanId)
                         </dd>
-                        @if (Model.IsPerUserPlan)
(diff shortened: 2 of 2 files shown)


src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
```diff
@@ -139,7 +139,6 @@ public IActionResult Index(string token = null)
                             this.subscriptionService.AddPlanDetailsForSubscription(subscriptionPlanDetail);
                             // GetSubscriptionBy SubscriptionId
                             var subscriptionData = this.apiClient.GetSubscriptionByIdAsync(newSubscription.SubscriptionId).ConfigureAwait(false).GetAwaiter().GetResult();
-                            subscriptionData.Quantity = 0;
```
dev-kiranbanda, line 142: Thank you- we were thinking that there should be a way to have the quantity captured on the landing page. This simplifies the process and doesn't require an extra step to do a **Change quantity** after the subscription is activated.
Besides, we should also try to have a min and max range enforced as it helps us handle the subscriptions against per user plans easily ( with validation in place).

I can approve this PR but please be informed that we are working on supporting the use case that I just explained. 

@code4clouds  -Please advise.

santhoshbomma9: Hi @dev-kiranbanda, 
I hope we are aligned here - perseat offer subscriptions already come with a user number in the getsubscriptionapi as the customers would have mentioned when they subscribe to a plan.  if we keep resetting the the flag here, database is never going to get the value from the api.
Change quantity is always required if the customer want to increase their user count during their subscription period
Sure, min, max should it should be good have the validations but again we have some new columns in the database for plan for min, max. And I also think we need save offerid in the subscription table as well as in the plan table, to load only plans relevant to that subscription. Will try to work on that.

dev-kiranbanda: @santhoshbomma9 - Would you mind sharing a screenshot of how the purchase experience looks like when customer tries to purchase a per-user plan? If the quantity is captured at the time of creating the resource itself, I am good with your change. I just want to be sure that there is an option to provide the number at the time of purchase.

santhoshbomma9: Sure @dev-kiranbanda , the below is where the customer would initially set the quantity when they are subscribing to an offer. and going forward they can update the quantity to their needs.
image

dev-kiranbanda: Thanks Santosh, this helps. Let me double check and approve the PR. We need to sync these changes and test at our end.

code4clouds:
@santhoshbomma9 Would you mind updating your PR to accommodate the new changes.

santhoshbomma9:
closing this one as its obselete. opening a new PR.

PR #60, opened 2020-05-13
Feature/2.1 library update newtonsoft

Indrasena-Bijjala-spektrasystems:
Replaced dependency on Newtonsoft Json with the System.Text.Json library. Updated the use of JSON Serialize and   Deserialize methods w.r.t System.Text.Json library


dev-kiranbanda:
> Replaced dependency on Newtonsoft Json with the System.Text.Json library. Updated the use of JSON Serialize and Deserialize methods w.r.t System.Text.Json library

I think we can abandon this PR as the PR related to base-use-case has taken care of this update.

dev-kiranbanda:
Closing this PR as the changes in this branch are considered in another PR that is active and would make it to master.

PR #82, opened 2020-06-11
ARM template to deploy web app and database

dev-kiranbanda:
One ARM template to deploy customer portal and publisher portal along with the database.


code4clouds:
@dev-kiranbanda I also found a good resource on how to create the app registration as part of the PowerShell: https://www.vrdmn.com/2017/05/create-azure-ad-app-registration-with.html

dev-kiranbanda:
> @dev-kiranbanda I think it will be better to remove the zip binary files and instead do:
> 
> ```
> dotnet build -c Release
> ```
> 
> Then zip the files as part of the PowerShell Compress-Archive command.
> 
> Once zipped continue with the process as you have done here.

I followed this and updated the powershell script to build, publish to local, push to storage account and use that path in the ARM template.
Removed the binaries from the repo.

PR #84, opened 2020-06-18
added github actions for ci/cd pipelines setup for both projects

santhoshb-msft:
added github actions for ci/cd pipelines setup for both projects.
By adding Publishprofile and AppName in GitHub secrets, we can have the CI/CD pipeline setup for both projects


santhoshb-msft:
> This is great! Lets go over on what account do we place this. Chat tomorrow?

Sounds good!

Issue #85, opened 2020-06-18
Plan IsPerUser value source of truth


santhoshb-msft:
**Is your feature request related to a problem? Please describe.**
Should the Plan IsPerUser dictated by the subscription details not by manually updating in the database.

**Describe the solution you'd like**
When the subscription details are loaded for the first time we might set this value

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

santhoshb-msft:
#86 updated

dev-kiranbanda:
I saw the way you derived this detail going by the quantity attribute on the subscription. However, it would have been nice, had there been an API to query for Plan detail. It helps us initialize the plan with lot more detail than what we have now.

santhoshb-msft:
List available plans https://learn.microsoft.com/en-us/azure/marketplace/partner-center-portal/pc-saas-fulfillment-subscription-api#get-httpsmarketplaceapimicrosoftcomapisaassubscriptionssubscriptionidlistavailableplansapi-versionapiversion API currently returns that information.

santhoshb-msft:
1. To update client lib to return all the info currently on the fuliflment api
2. update the ui logic based on the info which came back

PR #86, opened 2020-06-18
updated logic get isperuser plan from subscription

santhoshb-msft:
Isperuser plan value populated based on new subscription data
Issue #85


code4clouds:
@santhoshb-msft Would you rebase one more time?

PR #93, opened 2020-06-27
update arm to auto create app registration

santhoshb-msft:
Hi @code4clouds (Julio)
Keen on your thoughts. 
Changes
When Appregistration is passed empty
Create new Appregistration and Appsecret
Get URLS after ARM deployment
Assigned redirect urls to app registration
Add User.Read scope(used cli for this***)

### DIFF
--- a/deployment/Templates/Deploy.ps1
+++ b/deployment/Templates/Deploy.ps1
@@ -3,19 +3,19 @@
 #
 
 Param(  
-   [string]$WebAppNamePrefix, # Prefix used for creating web applications
-   [string]$TenantID, # The value should match the value provided for Active Directory TenantID in the Technical Configuration of the Transactable Offer in Partner Center
-   [string]$ADApplicationID, # The value should match the value provided for Active Directory Application ID in the Technical Configuration of the Transactable Offer in Partner Center
-   [string]$ADApplicationSecret, # Secret key of the AD Application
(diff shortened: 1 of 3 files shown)


deployment/Templates/Deploy.ps1
```diff
@@ -80,6 +80,22 @@ $PathToWebApplicationPackages = ((Get-AzStorageContainer -Container $ContainerNa
 
 Write-host "Path to web application packages $PathToWebApplicationPackages"
 
+if (!$ADApplicationID)
```
dev-kiranbanda, line 83: Thanks @santhoshb-msft for the changes to the powershell script. 
I think we should update the documentation to make it explicit that client ID is optional and that the script would create the AD app if none is provided as input.

Besides, we should also tell the user that this AD application ( if we are creating one through the PS ), should be used in the technical configuration section of the Offer in the Partner Center.

Please let me know your thoughts.

santhoshb-msft: Agree on both, _documentation_ and _should be used in the technical configuration section_.  will include them.

Keen on thoughts about the using az cli to assign the scope permissions to the app registration. So this means the publisher would need to install az cli and Az PS module which is what I'm kind of going back and forth.

santhoshb-msft:
Already implemented as part of a different PR.

PR #94, opened 2020-06-30
added logic to update isperuserplan when subscription gets created from landing page call

santhoshb-msft:
added logic to update isperuserplan when subscription gets created from landing page call


santhoshb-msft:
@code4clouds - given the use of client library, more changes are needed to make this work.

PR #111, opened 2021-04-13
Fixed Issue with Storage Blob.

code4clouds:
Fixed issue with blob dependency that breaks the deploy.ps1


code4clouds:
Fixed by #110

PR #132, opened 2021-09-29
fix for AAD delay, plus stop script if it fails

code4clouds:
Added a fix to wait for the AADID to be propagated as is not immediate. 
Also added a try/catch block to stop the script when this occurr.


dstarr (approved):
Looks good to manage failed resource allocation

PR #153, opened 2021-11-10
Add logic to inspect metered dimension

msalemcode:
Enhance the Plans update process to inspect metered dimension before inserting.
This will fix bug# 152


code4clouds (changes requested):
Same comments as PR#154

code4clouds:
Tested with old and new Landing Pages, but using the new Admin meter page changes.

code4clouds (approved):

PR #156, opened 2021-11-14
AZ CLI based deployment script.

alexdrenea:
Provides an updated script to deploy the entire solution with one single Az-Cli based script

The script can be ran from cloud shell so does not require the end-user to deploy / install anything on their local machine.

Notable changes from the original deploy.ps1 script:

- Fully AZ-CLI based, no dependencies on AzureRM
- Simplified mandatory parameter list with most of the parameters having defaults that are inferred from the chosen WebAppNamePrefix parameter
- SQL user / pass are auto-generated and discarded after the script runs for added security
- Added KeyVault integration with all secrets (SQL and AppRegistration secret) stored and referenced from a KeyVault for better security

### DIFF
--- a/deployment/Deployment/DeployCliFull.ps1
+++ b/deployment/Deployment/DeployCliFull.ps1
@@ -1,0 +1,484 @@
+﻿# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE file in the project root for license information.
+
+#
+# Powershell script to deploy the resources - Customer portal, Publisher portal and the Azure SQL Database
+#
+
+#.\DeployCliFull.ps1 -WebAppNamePrefix saasamp1 -Location canadacentral
+
+Param(  
+   [string][Parameter(Mandatory)]$WebAppNamePrefix, # Prefix used for creating web applications
+   [string][Parameter(Mandatory)]$Location, # Location of the resource group
+   [string][Parameter()]$ResourceGroupForDeployment, # Name of the resource group to deploy the resources - defaults to web app name prefix id not provided
+   [string][Parameter()]$TenantID, # The value should match the value provided for Active Directory TenantID in the Technical Configuration of the Transactable Offer in Partner Center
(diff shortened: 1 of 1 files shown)


dstarr:
We very much appreciate the work that has gone into this and we really like some of the choices you are making (KV, for example) in the install. We are hoping to use an all PowerShell installer script. Jumping back and forth between az CLI and PS within the same script can be very confusing for later maintenance.

A simple example is using `az command ...` followed by a PS `Write-Host` command. Another might be using curl instead of Invoke-RestMethod.

We're all-in on PowerShell and migrating that direction. Further, we want to use ARM for resource creation as much as possible as it is the standard template format for creating resources in Azure. Yes, they can be scripted, but that tends to lead to higher maintenance costs for the scripting.

Anxious to hear your thoughts.

alexdrenea:
Thank you David for your comments and for the opportunity to continue the conversation...

I would love to chat about the reasons of introducing Az Cli in the scripting even if it's not PowerShell native. 
Also would love to brainstorm the merits of replacing the ARM template with a pure CLI option.

Either way, my ultimate goal is to provide options for the end-users regardless of their preferences / skills are.

That being said, I'm open and willing to update the current deploy script to include some of the "goodies" from this update - open for discussion but I am thinking mostly of KeyVault, deploying without the temp storage account and simplified input parameter list.

dstarr:
Alex,

Now that we've had more conversations via chat, do you want to keep this PR open or work with a different one? It's been open for about 6 weeks, so it's likely behind a bit. Thgouhts?

alexdrenea:
Hey, sorry for the delay! Holidays and back to work have been busy times :)

I am working on a separate PR based on the conversations we've had - i.e. ARM + PowerShell. 
I still think there's value in a pure Az CLI script for customers who prefer CLI to ARM but if the team does not want this, happy to close the PR.

dstarr:
Thank you for acknowledging the goals of the project.

Please do.

PR #159, opened 2021-11-19
Enable Entity Framework Code-First Migrations for the database

### LINKED ISSUE #136: Allow new code updates without having to delete the db
**Is your feature request related to a problem? Please describe.**
As an IT administrator.,
I would like to update to the new code of the Accelerator
without having to delete the DB.

**Describe the solution you'd like**
If the script is re-run as the user if the db should be preserved (or even skipped) during an update

**Describe alternatives you've considered**

**Additional context**

alexdrenea:
This PR enables EF Core Migrations for a Code-First database deployment.

With this change, the database can be generated and seeded using the dotnet-ef update command after code deployment instead of importing a bacpac thus simplifying and accelerating initial deployment.

Furthermore, future changes to database schema can be easily managed and applied to the database without needing to re import the database. (fixes #136)


alexdrenea:
I think moving to EF migrations will help a lot with initial deployment speed and simplicity (no need to do a restore on deploy) as well as help with further development of the product as the schema evolves.

However, this being an old PR, not sure if the team is interested in moving in this direction so I will close for the moment as I do a refactoring on my fork.

If there is interest, happy to contribute it again.

muzzar78:
@alexdrenea I think the EF migrations idea is very good. I've been looking at how we can deploy the solution, tweak for our needs but then take updates as they are delivered to the main project. This would help a lot.

PR #160, opened 2021-11-19
ApplicationLogRepository async

### LINKED ISSUE #106: HomeController.Index() should be async and use await inside
`HomeController.Index()` is declared like this:

    public IActionResult Index(string token = null)

and has code inside that looks like this:

    var subscriptionPlanDetail = this.apiClient.GetAllPlansForSubscriptionAsync(newSubscription.SubscriptionId).ConfigureAwait(false).GetAwaiter().GetResult();

It cannot use `await` because it is not declared `async`. Other methods in the same class such as `ChangeSubscriptionPlan()` are declared async and use await no problem:

    public async Task<IActionResult> ChangeSubscriptionPlan(SubscriptionResult subscriptionDetail)

Having some methods `async` and using `await` and other methods non-async and using `GetResult()` makes this code unclear and hard to understand.

This should either be explained or changed such that `Index()` is declared `async` and uses `await`.

alexdrenea:
Repositories and Services in general could benefit from being async

Initial stage of async-"ifying" repositories and controllers throughout the app.

Fixes #106

### DIFF
--- a/src/SaaS.SDK.Client.DataAccess/Contracts/IApplicationLogRepository.cs
+++ b/src/SaaS.SDK.Client.DataAccess/Contracts/IApplicationLogRepository.cs
@@ -1,6 +1,7 @@
 ﻿namespace Microsoft.Marketplace.SaasKit.Client.DataAccess.Contracts
 {
     using System;
+    using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


code4clouds (approved):

code4clouds:
Thanks for the contributions to the SaaS Accelerator!!!

PR #182, opened 2022-01-16
Fix Webhook Controller (fix #179)

### LINKED ISSUE #179: Getting null Webhook Payload for Unsubscribe Action
**Describe the bug**
After activating a subscription, when I cancel the subscription from the portal, the webhook is getting called, but the payload is empty. The application log

**To Reproduce**
1. Subscribe to SaaS product
2. Click on link to Configure Account Now
3. Activate subscription from accelerator portal
4. Activate subscription via admin portal
5.  Cancel subscription from Azure portal as customer

**Expected behavior**
The Application log should have a record of the Unsubscribe Webhook action, and the subscription should show as Unsubscribed in the admin portal. The only entries I see in the Application Log are: "The Azure Webhook Triggered." These entries should be followed by the webhook payload JSON.

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser Chrome

alexdrenea:
- Controller POST action should be Task
- Remove Dispose from ApplicationLogRepository -> should not need to dispose manually.

Fixes #179

### DIFF
--- a/src/SaaS.SDK.Client.DataAccess/Services/ApplicationLogRepository.cs
+++ b/src/SaaS.SDK.Client.DataAccess/Services/ApplicationLogRepository.cs
@@ -18,27 +18,12 @@
         private readonly SaasKitContext context;
 
         /// <summary>
-        /// The disposed.
-        /// </summary>
-        private bool disposed = false;
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


msalemcode (approved):
reviewed the code and test it

Issue #170, opened 2021-12-09
Azure SQL firewall deployment problem - security vulnerability


alexdrenea:
The SQL database is being deployed with a firewall rule that allows ALL IPs to connect to the database thus rendering the default SQL firewall useless.

[deploy.json line 120-121](https://github.com/Azure/Commercial-Marketplace-SaaS-Accelerator/blob/a3810acb949250aec554cde29c9f89048cab3e13/deployment/Templates/deploy.json#L120-L121)

```
        "startIpAddress": "0.0.0.0",
        "endIpAddress": "255.255.255.255"
```

If the intent was to allow Azure IPs access to the SQL database, endIpAddress should also be "0.0.0.0" [docs](https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/servers/firewallrules?tabs=bicep#serverfirewallruleproperties)

santhoshb-msft:
Thanks @alexdrenea , this is fixed PR #244

Issue #95, opened 2020-07-04
'System.IO.DirectoryNotFoundException: Could not find a part of the path 'D:\home\site\wwwroot\runtimes\unix\'.


VikramVasudevan:
**Bug Description**
First of all, Thanks for the great application. Unfortunately I am unable to deploy successfully. The ARM template deployment step gives the following error.

.\Deploy.ps1  -WebAppNamePrefix "XXXXXXXX"    -TenantID "XXXXXX"    -ADApplicationID "XXXXX"    -ADApplicationSecret "XXXXX"    -SQLServerName "XXXXX"    -SQLAdminLogin "XXXXXX"    -SQLAdminLoginPassword "XXXXX"    -PublisherAdminUsers "XXXXXXX"    -BacpacUrl "someurl"    -ResourceGroupForDeployment "XXXXX"    -Location "East US"    -AzureSubscriptionID "XXXXX"    -PathToARMTemplate "\path\to\deploy.json" 

(actual values masked for security)

[...]

PR #183, opened 2022-01-17
Deployment updates

alexdrenea:
### Deployment script
- Simplifying Parameters for deployment script:
    - All SQL related parameters are optional and populated with defaults for simplicity and security.
    - Resource group name is optional and can be inferred from the web app prefix
    - ARM template default is the local ./deploy.json
- replace MSDeploy in ARM with PowerShell script Deploy
    - this fixes issue where cannot run the deployment script locally (#95)
    - simplifies deployment but removing the need to push zip packages to storage blob
    - clears the path for using Linux Web App in the future (the MSDeploy extension is only available for windows)
- Other updates
    - Replace `get-AzureRMContext` command with the newer AZ equivalent - `Get-AzContext` -> confirmed this works the same in the cloud shell
    - Updated method of creating the landingPageAppRegistration by using the `New-AzureADApplication` -> this removes the dependency on the `jq` utility that is not installed by default on users' machines

### ARM template
- Add KeyVault support
    - Arm template deploys a KeyVault
    - Web app get managed identity with access to KeyVault
    - Connection string to SQL and AAD secret get stored in KeyVault and referenced in Web App via App Service KeyVault References for optimal security
- Other
    - Fix SLQ firewall setting (#170)
    - Cleanup some variables and parameters in the ARM template.
    - Removed MSDeploy extension since it causes issues with publishing when not using correct `CompressArchive` version


dstarr (commented):
Not sure I am comfortable inferring the resource group name from the resource prefix. That seems like it would produce potentially unexpected behavior. @code4clouds what do you think?

alexdrenea:
That's fair. We can turn that back into a mandatory parameter. 
But it's still an option to provide it if users want to....Maybe we can look at adjusting the naming convention? I noticed a lot of folks (as well as the official MS docs) recommend pre-pending the Resource Group Name with "rg"

Open to make the change(s) either way

dstarr:
Nah, it's cool. Not that big a deal given it's optional. I'll try to run
the installer tomorrow.

deployment/Templates/deploy.json
```diff
@@ -79,29 +79,52 @@
...
     "appServicePlanName": "[concat(parameters('webAppNamePrefix'),'AmpSvcPlan')]",
     "ProvisioningServiceAppName": "[concat(parameters('webAppNamePrefix'),'-portal')]",
     "PublisherServiceAppName": "[concat(parameters('webAppNamePrefix'),'-admin')]",
-    "collation": "SQL_Latin1_General_CP1_CI_AS",
+    "KeyVaultName": "[toLower(concat(parameters('webAppNamePrefix'),'-kv'))]",
     "sqlDBName": "AMPSaaSDB"
   },
   "resources": [
+    {
+      "type": "Microsoft.KeyVault/vaults",
```
santhoshb-msft, line 91: Should KeyVault be optional?

code4clouds: Yes, KeyVault should be optional.

alexdrenea:
Happy to see this reviveved. Happy to finish the work on this if need be. Just let me know what needs to be done. 
@code4clouds , @msalemcode let me know if you want to book a chat to talk though this.

code4clouds:
We will follow this PR on #279

Issue #177, opened 2021-12-17
Cancel button on App Config Edit screen


dstarr:
**Is your feature request related to a problem? Please describe.**

When clicking "back" on my browser from the App Config edit page I get an error. As expected. But there is no other way to get back to the page without clicking "Save App Config."

**Describe the solution you'd like**
A simple **Cancel** button that takes me back to the main App Config Edit page.

**Describe alternatives you've considered**

A back button

**Additional context**

![MicrosoftTeams-image](https://user-images.githubusercontent.com/64742/146583852-a97435a9-4017-4757-ba4d-ad4fdd64efe6.png)

alexdrenea:
With the PR merged, I think this can be closed now!

PR #184, opened 2022-01-18
Add Back button to Application Config Details page

alexdrenea:
#177 

I considered naming the button "cancel" as suggested by the above mentioned ticket, but I noticed that the Save Action does not got back to the settings page either and user has to manually go back .

In this context, a back button might be more appropriate than a "cancel" as it could cause confusion if the user Saved before.


santhoshb-msft (approved):
Looks good!

code4clouds (approved):

Issue #166, opened 2021-12-04
PostUsage event returns status as null


cdhemant:
Hello,

My Azure managed application is in preview mode . So does it mean PostUsageEventAsync event will be successful but response will be null ??
{
		"status": null,
		"effectiveStartTime": null,
		"usageEventId": "00000000-0000-0000-0000-000000000000",
		"messageTime": "0001-01-01T00:00:00",
		"resourceId": "00000000-0000-0000-0000-000000000000",
		"quantity": 0,
		"dimension": null,
		"planId": null,
		"RequestID": null
}

Thanks
Hemant

alexdrenea:
This looks to be a bug in the code. 
I will provide a PR for the fix in a bit! 
Thank you for reporting the issue

PR #185, opened 2022-01-19
PostUsage event always returns null

alexdrenea:
Quick fix to return a proper MeteringUsageResult based on the response from the API rather than empty instance.

This addresses #166


dstarr (approved):
I see how simple this solution is. Very cool. 

I don't feel awesome about returning from inside a try/catch. That said, I'll accept this as is. Thanks for the PR!

dstarr:
Returns populated JSON from emitting the billing event.

Issue #196, opened 2022-02-23
Replace the Activate to Subscribe on Portal if SA is configured as Delay Activation


code4clouds:
**Describe the bug**
As a publisher,
I would like the SA to automatically detect delay activation and change the landing page "Activate" button to subscribe,
so, I don't have to update the code it and it's part of the out-of-the-box experience.

**To Reproduce**
1. Configure a SaaS Accelerator as delayed activation
1. Buy a subscription
2. on landing page notice the Activate button.

**Expected behavior**
Activate button should read "subscribe"

PR #201, opened 2022-02-28
Replace the Activate to Subscribe on Portal if SA is configured as Delay Activation #196

olsoro0:
Issue #196: Replace the Activate to Subscribe on Portal if SA is configured as Delay Activation

### DIFF
--- a/src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
+++ b/src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
@@ -243,6 +243,7 @@ namespace Microsoft.Marketplace.SaasKit.Client.Controllers
                             subscriptionExtension.CustomerEmailAddress = this.CurrentUserEmailAddress;
                             subscriptionExtension.CustomerName = this.CurrentUserName;
                             subscriptionExtension.SubscriptionParameters = this.subscriptionService.GetSubscriptionsParametersById(newSubscription.SubscriptionId, currentPlan.PlanGuid);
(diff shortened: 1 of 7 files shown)


src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
```diff
@@ -296,7 +297,7 @@ public IActionResult Subscriptions()
                     foreach (var subscription in subscriptionDetail.Subscriptions)
                     {
                         Plans planDetail = this.planRepository.GetById(subscription.PlanId);
-                        subscriptionDetail.IsAutomaticProvisioningSupported = Convert.ToBoolean(this.applicationConfigRepository.GetValueByName("IsAutomaticProvisioningSupported"));
+                        subscription.IsAutomaticProvisioningSupported = Convert.ToBoolean(this.applicationConfigRepository.GetValueByName("IsAutomaticProvisioningSupported"));
```
code4clouds, line 300: Why is this not subscriptionDetail as it used to be?  Line 486 you are adding this value again on subscriptionDetail.

olsoro0: subscriptionDetail on line 486 is of SubscriptionResultExtension type, while here subscriptionDetail was of SubscriptionViewModel type, which I've replaced with subscription of SubscriptionResultExtension type. In summary, in this PR, I've moved IsAutomaticProvisioningSupported property from SubscriptionViewModel to SubscriptionResultExtension, since we only use it on subscription level and this change was needed for the model in _LandingPage.cshtml.

code4clouds:
Thanks for the contribution!

code4clouds (approved):

PR #233, opened 2022-04-18
removed notstarted status

santhoshb-msft:
Removed unused NotStarted status.

### DIFF
--- a/src/SaaS.SDK.Services/Services/SubscriptionService.cs
+++ b/src/SaaS.SDK.Services/Services/SubscriptionService.cs
@@ -195,9 +195,8 @@ namespace Microsoft.Marketplace.SaaS.SDK.Services.Services
         /// <returns> Subscription Status EnumExtension.</returns>
         public SubscriptionStatusEnumExtension GetSubscriptionStatus(string subscriptionStatus)
         {
-            SubscriptionStatusEnumExtension status;
-            Enum.TryParse(subscriptionStatus, out status);
-            return status;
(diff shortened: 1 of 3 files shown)


code4clouds:
@santhoshb-msft  This may need a rebase..

src/SaaS.SDK.Services/Services/SubscriptionService.cs
```diff
@@ -195,7 +195,7 @@ public List<SubscriptionResult> GetPartnerSubscriptions(string partnerEmailAddre
         /// <returns> Subscription Status EnumExtension.</returns>
         public SubscriptionStatusEnumExtension GetSubscriptionStatus(string subscriptionStatus)
         {
-            var status = SubscriptionStatusEnumExtension.NotStarted;
+            SubscriptionStatusEnumExtension status;
```
code4clouds, line 198: This is missing the default value to return...

Try.Enum is usually inside an IF statement.  IF True returns the out, if False then dev needs to set the default value.   Ref the language spec: https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-6.0

code4clouds (changes requested):
Need change... on the Enum.Try

src/SaaS.SDK.Services/Services/SubscriptionService.cs
code4clouds: @santhoshb-msft ... any ETA on this change?

santhoshb-msft: Done! @code4clouds

code4clouds (approved):

PR #241, opened 2022-05-10
temp fix for db error when inserting plan description

santhoshb-msft:
temp fix for db error when inserting plan description


dstarr (approved):
This is a temporary worksound and we needed it for a live partner. Long term, we should likely remove the Plan Description.

PR #244, opened 2022-05-11
updated the ipfirewall rule for db

santhoshb-msft:
Update the db firewall to allow traffic from azureips only


code4clouds (approved):
Verified this and it allows private Ips

Issue #206, opened 2022-03-07
Fetch All subscriptions is assigned to the admin user


santhoshb-msft:
**Describe the bug**
Fetch All subscriptions is assigned to the admin user

**To Reproduce**

**Expected behavior**
Fetch All subscriptions is assigned to the beneficiary user. Check how the subscriptions are created from landingpage and the way the user is assigned. Use the same for fetch functionality on publisher portal as well

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #268, opened 2022-06-30
Fixed fetch all subscription issue (#206)

ms-mgrant:


src/SaaS.SDK.PublisherSolution/Controllers/HomeController.cs
```diff
@@ -834,6 +836,36 @@ public IActionResult FetchAllSubscriptions()
                             };
                             this.subscriptionLogRepository.Save(auditLog);  // add audit log
                         }
+                    } else
+                    {
```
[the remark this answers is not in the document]
ms-mgrant: This was only to fix the issue reactively, if the erroneous userId assignment was already made. Otherwise, the first block of code will run and the updated SubscriptionService should save the correct values.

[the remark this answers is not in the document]
santhoshb-msft: wouldn't that mean we are hiding/miss potentially erroneous code/logic somewhere and loss our opportunity to fix it. let's go with just only new subscriptions -> attach a user if existing, if not create user and attach to the sub.

src/SaaS.SDK.Client.DataAccess/Services/SubscriptionsRepository.cs
```diff
@@ -46,6 +46,8 @@ public int Save(Subscriptions subscriptionDetails)
                 existingSubscriptions.SubscriptionStatus = subscriptionDetails.SubscriptionStatus;
                 existingSubscriptions.AmpplanId = subscriptionDetails.AmpplanId;
                 existingSubscriptions.Ampquantity = subscriptionDetails.Ampquantity;
+                existingSubscriptions.UserId = subscriptionDetails.UserId;
```
santhoshb-msft, line 49: do we need this?

src/SaaS.SDK.PublisherSolution/Controllers/HomeController.cs
```diff
@@ -798,7 +798,8 @@ public IActionResult FetchAllSubscriptions()
                 var subscriptions = this.fulfillApiService.GetAllSubscriptionAsync().GetAwaiter().GetResult();
                 foreach (SubscriptionResult subscription in subscriptions)
                 {
-                    if(this.subscriptionRepo.GetById(subscription.Id) == null)
+                    // create subscription (and user) if it doesn't exist
```
santhoshb-msft, line 801: can you revert this, I dont think there is any logic change here right?

src/SaaS.SDK.Client.DataAccess/Services/SubscriptionsRepository.cs
ms-mgrant: Will remove, this was part of the sync functionality

santhoshb-msft (approved):
tested and looks good.

PR #271, opened 2022-07-06
Speed up subscriptions load time

ms-mgrant:


src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
```diff
@@ -300,9 +300,10 @@ public IActionResult Subscriptions()
                     this.TempData["ShowWelcomeScreen"] = "True";
                     SubscriptionViewModel subscriptionDetail = new SubscriptionViewModel();
                     subscriptionDetail.Subscriptions = this.subscriptionService.GetPartnerSubscription(this.CurrentUserEmailAddress, default, true).ToList();
+                    var allPlans = this.planRepository.Get().ToList();
```
santhoshb-msft, line 303: Can we not have the PlanId return when fetching subscriptions from the DB

ms-mgrant: If I restricted the allPlans to only fetch plans belonging to the user (in plan repo), would that solve the issue?

ms-mgrant: Would using the PlanGUId work better in this case?

code4clouds:
@ms-mgrant I would recommend that for new features that a screenshot of the changes is included. It just makes easy to understand the impact of the changes. ;-)

src/SaaS.SDK.CustomerProvisioning/Controllers/HomeController.cs
santhoshb-msft: True but we have to wait till the PlanGuidId change is merged.

PR #272, opened 2022-07-09
Fixed Unsubscribe in situation where Unsubscribed failed

k-granados:
This allows to unsubscribe when a subscription has a status of UnsubscribedFailed


santhoshb-msft (approved):
Tested changes. Approved.

PR #273, opened 2022-07-11
renamed css files to prevent caching

ms-mgrant:
This will fix the the issue of CSS files with the same name being cached, which causes the styling bug on site switch.


code4clouds (approved):

code4clouds:
Thanks for the code change.... Please resolve the conflicts  to merge the changes.  From the conflict message it seems you rename the files using the IDE instead of using the ```git mv``` command, which tend to cause these conflicts.

PR #277, opened 2022-07-15
Removed error message

louisephilardeau:
Removed error message on Manage Usage page.


code4clouds:
Hi @louisephilardeau  Would you mind sharing the reasoning for remving the error message?

louisephilardeau:
> Hi @louisephilardeau Would you mind sharing the reasoning for removing the error message?

Hi @code4clouds, @dstarr asked to remove the message, claiming that if there have been no usage events, that should be self-evident since they are not present.

code4clouds (approved):

Issue #236, opened 2022-04-26
Create a much unique name for temp storage in the script


santhoshb-msft:
**Describe the bug**

**To Reproduce**
1. Run the installation script
2. It failed with storage account name already used
3. $storagepostfix = Get-Random -Minimum 1 -Maximum 1000
$StorageAccountName = "amptmpstorage" + $storagepostfix       #enter storage account name

**Expected behavior**
Should run the script

**Screenshots**

**Environment (please complete the following information):**
 - Are you using the CloudShell?
 - If not using the CloudShell:
    - PowerShell version
    - Modules installed
    
**Additional context**

Issue #235, opened 2022-04-22
Stop install if there is no Tenant or Subscription entered.


code4clouds:
**Is your feature request related to a problem? Please describe.**
As a installer,
I would like the PowerShell script to block the install if there is no subscription or tenant entered in the prompts,
So I don't get a failed deployment when the application runs.

**Describe the solution you'd like**
On the prompt for the Subscription and Tenant ID... exit the install if either is blank.

**Describe alternatives you've considered**
Stay in a loop till the field is not empty

Issue #229, opened 2022-04-08
Deployment script doesn't capture default subscription


dstarr:
**Describe the bug**
When running the installer, and not passing in the -Subscription paramater, I am prompted for my subscription and to hit enter if I want to accept the default.

Unless I type in the subscription id, the script fails when trying to access a null subscription.

**To Reproduce**
1. Run the script without specifying a subscription
2. Do not enter a subscription when prompted

**Expected behavior**
The script selects the default description OR make tenant ID and subscription ID params only

Issue #119, opened 2021-09-13
Temp Resource Group


dstarr:
**Describe the solution you'd like**
Allow optional parameter for temp resource group

**Additional context**
May allow me to stop a collision with an existing resource group.

alexdrenea:
Could we do without the temp storage account in the first place?
I have a working update of the deploy script and ARM template that does not require the temp RG. The storage account is created in the main RG and only holds the bacpac, and can be used later for enabling logs or other features. Since it doesn't cost anything I see no issue leaving it deployed in the RG. 

Additionally, the deploy script is rewritten to use AZ CLI instead of AzureRM and adds KeyVault integration to the entire solution. If there is interest, I can quickly make a PR for it.

@dstarr @code4clouds

PR #279, opened 2022-07-20
Msalem/enhance deployment process

msalemcode:
1- Adding New deployment script to speed up the process by dropping ARM template dependency
1- Adding deployment script to deploy KV with SaaS Accelerator and secure the configuration secrets


deployment/Templates/Deploy.ps1
```diff
@@ -186,117 +186,86 @@ if($LogoURLico) {
...
-   bacpacUrl                    = "$URLToBacpacFromStorage"
-   SAASKeyForbacpac             = ""
-   PublisherAdminUsers          = "$PublisherAdminUsers"
-   PathToWebApplicationPackages = "$PathToWebApplicationPackages"
-   MeteredSchedulerSupport      = "$MeteredSchedulerSupport"
-}
-
-
 # Create RG if not exists
 New-AzResourceGroup -Name $ResourceGroupForDeployment -Location $location -Force
```
alexdrenea, line 219: we could be using az cli to deploy this as well. Since we're using it below to deploy the rest of resources.

deployment/Templates/Deploy.ps1
```diff
@@ -186,117 +186,86 @@ if($LogoURLico) {
...
+$WebAppNameService=$WebAppNamePrefix+"AmpSvcPlan"
+$WebAppNameAdmin=$WebAppNamePrefix+"-admin"
+$WebAppNamePortal=$WebAppNamePrefix+"-portal"
+$Connection="Data Source=tcp:"+$SQLServerName+".database.windows.net,1433;Initial Catalog=AMPSaaSDB;User Id="+$SQLAdminLogin+"@"+$SQLServerName+".database.windows.net;Password="+$SQLAdminLoginPassword+";"
+az appservice plan create -g $ResourceGroupForDeployment -n $WebAppNameService --sku B1
+
+
+Write-host "📜  Create publisher Admin webapp"
+az webapp create -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNameAdmin  --runtime dotnet:6
+az webapp config connection-string set -g $ResourceGroupForDeployment -n $WebAppNameAdmin -t mysql --settings DefaultConnection=$Connection
```
alexdrenea, line 250: Should the type be "SQLAzure" instead of mysql?

alexdrenea:
Love seeing this making it into the accelerator. Hope to see this approved soon!

deployment/Templates/Deploy.ps1
```diff
@@ -186,117 +186,86 @@ if($LogoURLico) {
...
-New-AzStorageContainer -Name $ContainerName -Context $ctx -Permission Blob 
-Set-AzStorageBlobContent -File $LocalPathToBacpacFile -Container $ContainerName -Blob $BlobName -Context $ctx -Force
-
-$URLToBacpacFromStorage = (Get-AzStorageBlob -blob $BlobName -Container $ContainerName -Context $ctx).ICloudBlob.uri.AbsoluteUri
-
-Write-host "📜  Uploaded the bacpac file to $URLToBacpacFromStorage"    
-
+Write-host "☁  Prepare publish files for the web application"
 
 if (!($MeteredSchedulerSupport))
```
alexdrenea, line 191: This parameter could be a Boolean instead of string
`[bool][Parameter()]$MeteredSchedulerSupport`

and then it can be queried with 
`if ($MeteredSchedulerSupport -eq $true)`

When calling the script, set this to true by just specifying it in the command (-MeteredSchedulerSupport) or have it as false by omitting it.

deployment/Templates/Deploy.ps1
```diff
@@ -186,117 +186,86 @@ if($LogoURLico) {
...
-
 # Create RG if not exists
 New-AzResourceGroup -Name $ResourceGroupForDeployment -Location $location -Force
 
 Write-host "📜  Deploying the ARM template to set up resources"
-# Deploy resources using ARM template
-New-AzResourceGroupDeployment -ResourceGroupName $ResourceGroupForDeployment -TemplateFile $PathToARMTemplate -TemplateParameterObject $ARMTemplateParams
 
+Write-host "Create SQL Server"
+az sql server create --name $SQLServerName --resource-group $ResourceGroupForDeployment --location "$location" --admin-user $SQLAdminLogin --admin-password $SQLAdminLoginPassword
```
alexdrenea, line 224: The outputs from these commands can be quite verbose and will pollute the output window making it hard to follow. I usually append `--output none` to all the az cli commands to keep the output clean.

deployment/Templates/DeployKV.ps1
```diff
@@ -0,0 +1,315 @@
...
+# Create RG if not exists
+New-AzResourceGroup -Name $ResourceGroupForDeployment -Location $location -Force
+
+Write-host "📜  Start Deploy resources"
+$WebAppNameService=$WebAppNamePrefix+"AmpSvcPlan"
+$WebAppNameAdmin=$WebAppNamePrefix+"-admin"
+$WebAppNamePortal=$WebAppNamePrefix+"-portal"
+$KeyVault=$WebAppNamePrefix+"-kv"
+$ADApplicationSecretKeyVault="@Microsoft.KeyVault(SecretUri=https://"+$KeyVault+".vault.azure.net/secrets/ADApplicationSecret/)"
+$DefaultConnectionKeyVault="@Microsoft.KeyVault(SecretUri=https://"+$KeyVault+".vault.azure.net/secrets/DefaultConnection/)"
```
alexdrenea, line 227: Due to a bug (?) in either powershell or az cli settings command this style of declaring the key vault reference will not work. (if you look in the app settings config, the value won't be recognized as a keyvault referece as it is missing the closing ')'.

The below method should work instead
```
$ADApplicationSecretKeyVault='"@Microsoft.KeyVault(VaultName={0};SecretName=ADApplicationSecret)"' -f $KeyVault
$DefaultConnectionKeyVault='"@Microsoft.KeyVault(VaultName={0};SecretName=DefaultConnection)"' -f $KeyVault
```

deployment/Templates/DeployKV.ps1
```diff
@@ -0,0 +1,315 @@
...
+$Connection="Data Source=tcp:"+$SQLServerName+".database.windows.net,1433;Initial Catalog=AMPSaaSDB;User Id="+$SQLAdminLogin+"@"+$SQLServerName+".database.windows.net;Password="+$SQLAdminLoginPassword+";"
+
+
+Write-host "Create SQL Server"
+az sql server create --name $SQLServerName --resource-group $ResourceGroupForDeployment --location "$location" --admin-user $SQLAdminLogin --admin-password $SQLAdminLoginPassword
+
+Write-host "Add SQL Server Firewall rules"
+az sql server firewall-rule create --resource-group $ResourceGroupForDeployment --server $SQLServerName -n AllowYourIp --start-ip-address "0.0.0.0" --end-ip-address "0.0.0.0"
+
```
alexdrenea, line 236: When running this locally, the script will fail because the IP of the machine running the script is not part of the firewall.
We could solve this by adding local user's IP to the firewall.
The below works, but happy to try out other things that don't depend on a 3rd party website.

```
$publicIp = (Invoke-WebRequest -uri "http://ifconfig.me/ip").Content
az sql server firewall-rule create --resource-group $ResourceGroupForDeployment --server $SQLServerName -n AllowYourIp --start-ip-address "$publicIp" --end-ip-address "$publicIp"
```

deployment/Templates/DeployKV.ps1
```diff
@@ -0,0 +1,315 @@
...
+$ADApplicationSecretKeyVault="@Microsoft.KeyVault(SecretUri=https://"+$KeyVault+".vault.azure.net/secrets/ADApplicationSecret/)"
+$DefaultConnectionKeyVault="@Microsoft.KeyVault(SecretUri=https://"+$KeyVault+".vault.azure.net/secrets/DefaultConnection/)"
+$Connection="Data Source=tcp:"+$SQLServerName+".database.windows.net,1433;Initial Catalog=AMPSaaSDB;User Id="+$SQLAdminLogin+"@"+$SQLServerName+".database.windows.net;Password="+$SQLAdminLoginPassword+";"
+
+
+Write-host "Create SQL Server"
+az sql server create --name $SQLServerName --resource-group $ResourceGroupForDeployment --location "$location" --admin-user $SQLAdminLogin --admin-password $SQLAdminLoginPassword
+
+Write-host "Add SQL Server Firewall rules"
+az sql server firewall-rule create --resource-group $ResourceGroupForDeployment --server $SQLServerName -n AllowYourIp --start-ip-address "0.0.0.0" --end-ip-address "0.0.0.0"
```
alexdrenea, line 235: [Nit] The rule name should be "AllowAzureIP" not "Your" Ip since this rule is essentially the convention for allowing other Azure services to access this SQL server

deployment/Templates/Deploy.ps1
msalemcode: Added the recommendation

deployment/Templates/Deploy.ps1
msalemcode: Not sure. It make should be SQLAzure type however it was always mysql.

deployment/Templates/DeployKV.ps1
msalemcode: we ran this code mutliple time and I re-run it again after reading your comment and still I am not getting any error. after deployment web app is up, It authenticate users and it bring data back from database. it is indication the KV reference is working. Did you try latest code

deployment/Templates/DeployKV.ps1
msalemcode: In case of running locally. User should modify the rule based on their needs. we have ISV who running it from CI/CD pipeline. I will leave this rule as it is for now and local user should take control over how they want to modify the DB rules.

deployment/Templates/DeployKV.ps1
msalemcode: updated the rule name

deployment/Templates/Deploy.ps1
msalemcode: when we run into errors during the deployment this verbose does help to track down the error location

code4clouds:
@msalemcode  @alexdrenea  Is this good to go?  Should we close PR #183 ?

deployment/Templates/DeployKV.ps1
```diff
@@ -0,0 +1,331 @@
...
+$ErrorActionPreference = "Stop"
+
+# Checking SQL username
+if($SQLAdminLogin.ToLower() -eq "admin") {
+    Throw "🛑 SQLAdminLogin may not be 'admin'."
+    Exit
+}
+
+# Checking SQL password length
+if($SQLAdminLogin.Length -lt 8) {
```
alexdrenea, line 35: this should be SqlAdminPassword

deployment/Templates/DeployKV.ps1
alexdrenea: this could be due to the PowerShell version. 
It certainly does not work in PowerShell 5.1 (and does not really throw an error, it will simply not work after deployment). while this version might work in PowerShell 7, the proposed one will work in either version.

Happy to run a test together and make sure we land on a generic solution that works everywhere.

code4clouds: The important version is the one in the Azure Portal cloud shell...  as that is the target version for the install instructions, all other are best effort.

alexdrenea: @code4clouds  I understand that, but in my experience, a lot of partners I worked with want to be able to clone and deploy from local. Windows 10 and 11 come preinstalled with PowerShell 5.1 and honestly the script works perfectly with this small exception. I don't see a reason why we could not make the script work on both Ps version :)

alexdrenea:
Reviewed with Magdy today! All looks good! 

Unfortunately it looks like I don't have permissions  / access to formally Approve the review, but +1 from my end!

code4clouds:
@alexdrenea Try again...  I granted triage access. :-)

alexdrenea (approved):

alexdrenea:
@code4clouds I approved!

I can't link issues to the PR but I think this could also solve:
#88 -> KeyVault is added by default in the deployment, no longer needs extra instructions
#95 -> moving the deployment out of the ARM template resolves the compress-archive issue in PowerShell 5.1
#119 -> temp Resource Group is no longer needed
#229, #235 -> the script uses the new Azure PowerShell commands, and confirmed to work both locally and in CloudShell without entering SubId / TenantId
#236, #293 -> the new script does not create temp storage accounts

code4clouds (changes requested):
Please update the changes requested ASAP to start the final testing.

deployment/Templates/DeployKV.ps1
```diff
@@ -0,0 +1,331 @@
+﻿# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
alexdrenea: I support this as well. 
I am of strong opinion that KeyVault should not be optional, and think we can remove the "non-KeyVault" file entirely and just have the one deployment file that comes with KeyVault.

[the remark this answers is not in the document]
code4clouds: +1 let's remove the non-keyvault file and just have one called Deploy.ps1

code4clouds (approved):

code4clouds (approved):

code4clouds (approved):

PR #282, opened 2022-07-25
Fix for password check

alexdrenea:
Correct check for SQL admin pass.


santhoshb-msft:
### **Ran script with two char password - Fails during execution**

image
image


### **Ran script with current change and two char password - Stops the execution**

image

### Ran script with current change with proper password - Execution successful

image
image

santhoshb-msft (approved):
tested changes. Looks good

code4clouds (approved):
Thanks for this update.

Issue #281, opened 2022-07-21
Display beneficiary information in the subscription details 


santhoshb-msft:
**Is your feature request related to a problem? Please describe.**
Display beneficiary information in the subscription details 

**Describe the solution you'd like**
Able to see beneficiary information

**Describe alternatives you've considered**

**Additional context**

santhoshb-msft:
Merged PR linked

PR #284, opened 2022-07-26
Added beneficiary details to subscription details

ms-mgrant:
Issue #281 

Screenshot of changes:
image


santhoshbomma9:
@ms-mgrant  - I think should save and pull this from DB.

ms-mgrant:
These fields aren't available in the DB. Is it worth making the modifications to add them instead of using the API?

code4clouds (approved):

PR #288, opened 2022-07-27
added performance fixes for subscriptions

ms-mgrant:
This solution reduces the admin subscriptions load time by ~75% and the customer subscription load time by ~45%.

Benchmarks on local device (averages):
- Admin page load time went from 2:05 to around 30 seconds.
- Customer page went from 1:15 to around 40 seconds.


santhoshb-msft:
Closing this PR as it will handled in PR #304 which is built on top of this PR.

PR #289, opened 2022-07-28
implement fetch all fix

ms-mgrant:
Creates a workaround for saving subscriptions with an "unsubscribed" status within FetchAllSubs()


santhoshb-msft:
Issue fixed in PR #305

PR #294, opened 2022-08-05
Added webhook renew event

dstarr:
When a Renewed event is received on the webhook, a message is written to the event log. The event is no longer throwing an error or being swallowed.


santhoshb-msft (approved):
image

 New and existing notification working fine. All good from me.

dstarr:
Had a mispelling in RenewedAsync, plus added some comments

PR #270, opened 2022-07-06
Show Offer name on Admin Page Issue #218

k-granados:
Added Offer Name column with the associated offer names in the subscriptions page on the Publisher Solution.

PR #301, opened 2022-08-11
Show Offer name on Admin Page Issue 

ms-mgrant:
Copy of PR #270 by @k-granados without merge conflicts.


santhoshb-msft:
image
There is data in the backup, subscription/offer/plan table,

deployment/Database/AMP-DB-5.1.sql
```diff
@@ -385,6 +385,7 @@ CREATE TABLE [dbo].[Subscriptions](
 	[AMPQuantity] [int] NOT NULL,
 	[PurchaserEmail] [varchar](225) NULL,
 	[PurchaserTenantId] [uniqueidentifier] NULL,
+    [PlanGUId] [uniqueidentifier] NULL
```
alexdrenea, line 388: Do we need this new field in the DB?

It doesn't look like it's needed for showing the offer name on the admin page

santhoshb-msft:
@alexdrenea I will close this PR and open a new one.
To answer your question, yes we need this incase there are two planId with the sane name. Its hard to distinguish which planId the sub is referring to. Ideally we should have foreign key relationship between , so weighing up the options.

deployment/Database/AMP-DB-5.1.sql
santhoshb-msft: @alexdrenea I will close this PR and open a new one.
To answer your question, yes we need this incase there are two planId with the sane name. Its hard to distinguish which planId the sub is referring to. Ideally we should have foreign key relationship between , so weighing up the options.

santhoshb-msft:
Will open as a new PR

PR #312, opened 2022-08-29
fixed typo

santhoshb-msft:
fixed typo in appsettings


dstarr:
Thank you for picking that up, Santhosh.

dstarr:
Thank you for picking that up, Santhosh.

PR #324, opened 2022-09-27
added logic to handle onetime payment schedule

santhoshb-msft:
added logic to handle onetime payment schedule


santhoshb-msft:
Need to update DB file

deployment/Templates/Deploy.ps1
```diff
@@ -198,13 +198,9 @@ Write-host "☁  Prepare publish files for the web application"
 Write-host "☁  Preparing the publish files for PublisherPortal"  
 dotnet publish ..\..\src\SaaS.SDK.PublisherSolution\SaaS.SDK.PublisherSolution.csproj -c debug -o ..\..\Publish\PublisherPortal
 
-if ($MeteredSchedulerSupport -ne $true)
```
code4clouds, line 201: Are we making the meter scheduler permanent since we are removing this option?

deployment/Database/AMP-DB.sql
```diff
@@ -851,23 +851,10 @@ BEGIN
 END
 GO
 
-IF NOT EXISTS (SELECT * FROM [dbo].[SchedulerFrequency] WHERE [Frequency] = 'Hourly')
```
code4clouds, line 854: If you are removing these frequencies, check with msalem to see what will be the impact.

code4clouds (changes requested):
"added logic to handle onetime payment schedule" - It seems you remove logic of current features, instead of just adding.

deployment/Templates/Deploy.ps1
santhoshb-msft: I believe that is correct, @msalemcode do you want to comment here.

PR #332, opened 2022-10-21
Dstarr fixing docs

dstarr:
Just removed a last couple of references to "SaaS.SDK." in READMEs.


santhoshb-msft (approved):
All wiki changes looks good. approved

PR #339, opened 2022-10-27
update term unit to cover 2Y and 3Y

santhoshb-msft:
update term unit to cover 2Y and 3Y


dstarr (approved):
Tested the change on a 3-year subscription and all works as expected.

Issue #345, opened 2022-11-03
Plan name does not change in Admin portal


alexdrenea:
**Describe the bug**
When changing plans (either customer initiated or partner initiated) the plan name is not updated in the admin portal subscriptions page or details.
This can be confusing and worrying for the partner as they can't tell if their change has taken effect or not.

**To Reproduce**
Change plan for one of the subscriptions 

**Expected behavior**
Once change plan is complete, I would like to see the new plan name in the offers page.

**Screenshots**
![image](https://user-images.githubusercontent.com/8275679/199807777-e7459cf9-de3e-4e06-a155-dc6bb7f99b10.png)

**Additional Context**
Ideally the change would happen automatically once the plan is changed, however, it would be acceptable if this would work when clicking "Fetch all Subscriptions" button

alexdrenea:
@santhoshb-msft - I can work on this if you don't mind. I have a branch that I started for this. no need to duplicate efforts

santhoshb-msft:
Sounds good @alexdrenea !

santhoshb-msft:
But there is the webhook flow I would like to mention, lets connect offline.

santhoshb-msft:
@alexdrenea,  can you help with two things please
1. On fetch lets sync all three when different, like state/plan/quantity
2. and add an audit log as well

Issue #219, opened 2022-03-21
Option to sync status/plan/quantity from the api on demand


santhoshb-msft:
**Is your feature request related to a problem? Please describe.**
Option to sync status/plan/quantity from the api on demand

**Describe the solution you'd like**
Currently I dont have a way to sync the status from the api on demand.

**Describe alternatives you've considered**
A button Sync status to fetch all subscriptions but also sync the status/plan/quantity

**Additional context**

PR #349, opened 2022-11-07
Refresh plan name

### LINKED ISSUE #353: Wrong SubscriptionStatus in Subscription Details page
**Describe the bug**
In Admin Portal, Subscription Details page displays wrong Status (it is hooked up to SaaSSubscriptionStatus property instead of SubscriptionStatus so it always shows PendingFulfullmentStart

**To Reproduce**
1. Go to Admin Portal
2. Click on 'Subscription Details"

**Expected behavior**
Show correct Status as presented in the subscriptions table.

alexdrenea:
### Logic:

- when changing plan, ensure we update the database with the new plan for the offer (and log)
- when fetching subscription list, ensure we refresh plan name, quantity or status and log

### UI:
- change plan / change quantity are long running operations (30s-60s). Disable button and show spinner on page while operation is in progress for both change plan and change quantity modals both in admin portal and customer portal
 
image

Addresses #345 , #350
Also fixes #353 , #354


santhoshb-msft:
@alexdrenea I believe this should always be handled by the response on the webhook. I think we can do it here too but webhook is generally the better place to handle this update.

alexdrenea:
> @alexdrenea I believe this should always be handled by the response on the webhook. I think we can do it here too but webhook is generally the better place to handle this update.

I like this! Lemme look into it!
Meanwhile, if you get a moment to check the rest of the PR that would be great!

alexdrenea:
> @alexdrenea I believe this should always be handled by the response on the webhook. I think we can do it here too but webhook is generally the better place to handle this update.

@santhoshb-msft I verified, and it looks like the change is applied correctly in the webhook as you mentioned. I could have sworn I saw an issue where the plan was not updated. But I retried it both from customer portal and admin portal and it works as expected.

And in any case, if there is any kind of transient problem, the "Fetch All Subscription" button will now force refresh everything.

Thank you. I reverted all the changes in HomeController related to changing the plan there!

santhoshb-msft:
@dstarr this might be fixing #219 as well

dstarr (commented):
This is not working for me at present. The UX looks like it runs properly, but the plan is not changed in the Subscriptions table.

The plan IS changed however when viewed from the customer's view in the Azure portal.

dstarr (changes requested):
The changes to **FetchAllSubscriptions** are not synching Plans or Offers. The call completes with a 200, but only the Subscriptions table is populated.

PR #364, opened 2022-12-01
Modernize Deployment script

### LINKED ISSUE #359: Modernize deployment script
Our deployment script that is foundational to the easy deployment of the Accelerator is one of the most important value-add to our partners.

### Tasks:
1. **Use `az ad app` command set for managing the app registrations**
    While parts of the script have been recently modernized to take advantage of new enhancements in Azure's deployment framework (`az cli`), the app registration creation is still using older PowerShell cmdlets.
    **Benefits:**
    - take advantage of latest improvements in creating app registrations
    - eliminate the hardcoded JSON payload that is currently sent for the Landing Page App Reg (cleaner safer code, easier to maintain)
    - az ad app commands are idempotent unlike the old ones which always create new app registrations (in case of failures when you need to run scripts multiple times)
    - new features enable us to upload the logo for the landing page app branding
    - eliminate dependency on `jq` 
1. **Standardize and clean up script output.** 
    As you look at the current output It looks like as features were added to the script there was never consistency in how and what messages we log. Also, while helpful for debugging, the command outputs conflicts and clashes with the script output making it hard to go through.
    **Benefits**
    - clean and polished logs make for a nicer deployment experience
    - add a "debug" flag to enable / disable az cli script outputs allows for both clean ouptut and debugging when needed
Example of potential script log output:
image

alexdrenea:
1. **Use `az ad app` command set for managing the app registrations**
    While parts of the script have been recently modernized to take advantage of new enhancements in Azure's deployment framework (`az cli`), the app registration creation is still using older PowerShell cmdlets.
    **Benefits:**
    - take advantage of latest improvements in creating app registrations
    - eliminate the hardcoded JSON payload that is currently sent for the Landing Page App Reg (cleaner safer code, easier to maintain)
    - az ad app commands are idempotent unlike the old ones which always create new app registrations (in case of failures when you need to run scripts multiple times)
    - new features enable us to upload the logo for the landing page app branding
    - eliminate dependency on `jq` 
      image

1. **Standardize and clean up script output.** 
    As you look at the current output It looks like as features were added to the script there was never consistency in how and what messages we log. Also, while helpful for debugging, the command outputs conflicts and clashes with the script output making it hard to go through.
    **Benefits**
    - clean and polished logs make for a nicer deployment experience
    - add a "debug" flag to enable / disable az cli script outputs allows for both clean ouptut and debugging when needed
Script Output:
image

Fixes #359

### DIFF
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -5,8 +5,16 @@
 # Powershell script to deploy the resources - Customer portal, Publisher portal and the Azure SQL Database
 #
 
+#.\Deploy.ps1 `
+# -WebAppNamePrefix "amp_saas_accelerator_<unique>" `
+# -Location "<region>" `
+# -PublisherAdminUsers "<your@email.address>"
+
 Param(  
    [string][Parameter(Mandatory)]$WebAppNamePrefix, # Prefix used for creating web applications
+   [string][Parameter()]$ResourceGroupForDeployment, # Name of the resource group to deploy the resources
+   [string][Parameter(Mandatory)]$Location, # Location of the resource group
+   [string][Parameter(Mandatory)]$PublisherAdminUsers, # Provide a list of email addresses (as comma-separated-values) that should be granted access to the Publisher Portal
    [string][Parameter()]$TenantID, # The value should match the value provided for Active Directory TenantID in the Technical Configuration of the Transactable Offer in Partner Center
    [string][Parameter()]$AzureSubscriptionID, # Subscription where the resources be deployed
    [string][Parameter()]$ADApplicationID, # The value should match the value provided for Active Directory Application ID in the Technical Configuration of the Transactable Offer in Partner Center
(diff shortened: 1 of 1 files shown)


dstarr (commented):
Please also update the installation.md instructions to include the new flag.

dstarr:
Bug fix in #391 means you need to clear a merge conflict, unfortunately. No biggie, I would think.

deployment/Deploy.ps1
```diff
@@ -56,173 +74,164 @@ if($WebAppNamePrefix.Length -gt 21) {
...
-
-        # Download Publisher's AppRegistration logo
+		$landingpageLoginAppReg = `
+			az ad app create `
+				--only-show-errors `
+				--display-name "$WebAppNamePrefix-LandingPageAppReg" `
+				--sign-in-audience "AzureADandPersonalMicrosoftAccount" `
+				--web-redirect-uris `
+					https://$WebAppNamePrefix-portal.azurewebsites.net `
+					https://$WebAppNamePrefix-portal.azurewebsites.net/ `
```
santhoshb-msft, line 169: @alexdrenea are we missing the logout URL here?

santhoshb-msft (changes requested):
Added comment on the logout URL to multi-tenant appid?

deployment/Deploy.ps1
```diff
@@ -15,38 +23,48 @@ Param(
...
+# Install-Module Az
+# Install-Module -Name AzureAD
+
 $ErrorActionPreference = "Stop"
-$SaaSApiConfiguration_CodeHash= git log --format='%H' -1
-#Setting up default Sql Parameters
+
+#region Set up Variables and Default Parameters
+
+if ($ResourceGroupForDeployment -eq "") {
```
msalemcode, line 41: Please change the logic to check if the parameter exists. I executed the script without providing SQL username/password with expectation the code will generate one for me. 
image

Received Error **InvalidParameters** 
image

Line 308 error out
image

deployment/Deploy.ps1
alexdrenea: I don't think that was ever used. 
Our app doesn't even expose a /logout endpoint.

santhoshb-msft: could you please add either way? I think we need to fix that right? Current logout seems to throw error.

santhoshb-msft: image

alexdrenea: @santhoshb-msft I still don't think that the setting is necessary or even correct.

If you look at the value of the logout url, that points to "https://<yourapp>-admin.azurewebsites.net/logout". The '/logout' controller path does not exist in our app and thus throws a 404. As such, setting that value in our app registration will do no good.

The current bug recorded against the SignOut controller is unrelated given that this PR is not merged yet (see comment on the bug)

santhoshb-msft: @alexdrenea - my side point is that whne its time to fix this bug and set the logout url, I dont believe you can do that other than with az rest. so there is no point moving to az ad app create on MT app reg.

dstarr: It's ridiculous, but true.

alexdrenea: It is doable with `az ad app` - it's actually pretty simple. If you insist I can add it, but we don't actually have a proper value to set it to... what is currently used is incorrect and should be removed anyway (there is no /logout endpoint exposed in our app)

alexdrenea: Agree it's a bit annoying. 
But have we established if we even need the logoutUrl? I've tested the app and have not seen any difference in behavior (again, even today, the fact that we have it set to a wrong value does not really make a difference)
Also, if we ever find a use case where we do need it, it can be added with an additional `az rest` command with a much smaller and manageable payload (or by that time maybe it will be added in the `az ad app` command parameter set)
I still think creating the app with `az ad app` is beneficial as it handles setting the majority of parameters we need as simple arguments rather than setting them in a json payload.

alexdrenea:
going to close this PR due to merge conflicts (I messed up the merge in my fork). Will re-open a new one and continue conversation there.

PR #365, opened 2022-12-01
Fix compiler warnings

### LINKED ISSUE #363: Fix compilation warnings that show up when deploying solution
**Is your feature request related to a problem? Please describe.**
This is not a blocking issue, but every time we run the deployment script, we see 6 warnings when compiling and building packages for the Admin Portal and Customer portal.
We should fix the warnings.

**Describe the solution you'd like**
Fix the warnings

**Describe alternatives you've considered**
We could suppress the warnings but fixing them would be clearly better :)

**Additional context**
image

alexdrenea:
Fixes #363

### DIFF
--- a/src/DataAccess/Services/PlansRepository.cs
+++ b/src/DataAccess/Services/PlansRepository.cs
@@ -78,7 +78,7 @@
                     existingPlan.DisplayName = planDetails.DisplayName;
                     existingPlan.OfferId = planDetails.OfferId;
                     existingPlan.IsmeteringSupported = planDetails.IsmeteringSupported;
-                    this.CheckMeteredDimension(planDetails,existingPlan);
+                    this.CheckMeteredDimension(planDetails, existingPlan);
                     this.context.Plans.Update(existingPlan);
                     this.context.SaveChanges();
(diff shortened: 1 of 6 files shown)


src/DataAccess/Services/PlansRepository.cs
```diff
@@ -214,34 +214,28 @@ public PlanAttributeMapping GetPlanAttributeOnOfferAttributeId(int offerAttribut
         /// <returns> List type="of Plan attributes.</returns>
         public IEnumerable<PlanAttributesModel> GetPlanAttributes(Guid planGuId, Guid offerId)
         {
-            try
```
dstarr, line 217: Did you remove the try/catch because you are passing to the base controller?

dstarr (commented):
Left one small question for now.

src/DataAccess/Services/PlansRepository.cs
dstarr: Nevermind, it was because it was re-throwing. Got it.

dstarr (approved):
All changes tested and compiler warnings are gone.

PR #366, opened 2022-12-01
Ds namespace refactoring

dstarr:
I did not test the migrations. Do I do this by running an UPDATE install script?

This is a simple refactoring of namespaces.

1. Namespaces match file paths
2. Namespaces are now file scoped, except for autogenerated migration files, which retain their namespace structure.


alexdrenea:
@dstarr migrations also run on the deployment script (essentially, instead of our hardcoded sql script for deploying database, we ask Entity Framework to generate the script based on the migrations scripts that are also auto-generated)

The update script does almost the same thing as the deploy script except infrastructure (i.e. deploys code and database changes only).

So, if the deployment script works with the new namespace changes, it means everything including migrations work correctly.

Looking at the changes, I see no reason for it to fail.

alexdrenea:
There are 4 files under the "Migrations" folders that were changed. 
Those files are auto-generated by Entity Framework Migrations tool so they will probably get changed again next time we make a DB change. 
Could we revert them and leave them as-is?

dstarr:
Closing this pull request until other PRs are accepted and code-freeze is implemented.

PR #395, opened 2022-12-09

nickward13:
added a note to instruct user to use PowerShell instead of bash shell within Azure Cloud Shell


santhoshb-msft (approved):
Thank you for this, @nickward13

PR #397, opened 2022-12-10
Should not override the SignOut BaseController method.

### LINKED ISSUE #396: Admin logout not working
**Describe the bug**
Login seems to fails with no controllers

**To Reproduce**
1. Login to the app
2. Click on Logout
3. See error

**Expected behavior**
Should be able to logout without error and take back to login back

**Screenshots**
image

**Environment (please complete the following information):**
 - Are you using the CloudShell?
 - If not using the CloudShell:
    - PowerShell version
    - Modules installed
    
**Additional context**

alexdrenea:
The SignOut method in the BaseController is a NonAction Helper method. Our controller's SignOut needs to be "new" as it's an action we want to call (arguably, should not have called it SignOut)

fixed #396

### DIFF
--- a/src/AdminSite/Controllers/AccountController.cs
+++ b/src/AdminSite/Controllers/AccountController.cs
@@ -29,7 +29,7 @@
         /// <returns>
         /// The <see cref="IActionResult" />.
         /// </returns>
-        public override SignOutResult SignOut()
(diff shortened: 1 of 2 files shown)


alexdrenea:
@santhoshb-msft  @dstarr  minor PR to fix signout issue.

Issue #354, opened 2022-11-10
Can't change Quantity in Admin Portal


alexdrenea:
**Describe the bug**
The Change Quantity Popup does not load correctly in the Admin Portal

**To Reproduce**
1. Go to Admin Portal
2. Click on 'change quantity"

**Expected behavior**
Able to change quantity 

**Screenshots**
![image](https://user-images.githubusercontent.com/8275679/200973969-688fb495-ddc6-41c3-9d22-a2398d5b8a80.png)

**Additional context**
This is likely due to using an invalid SubscriptionService Method to fetch the subscription Details in `HomeController` / `SubscriptionQuantityDetail` method (should use `subscriptionService.GetSubscriptionsBySubscriptionId`)

santhoshb-msft:
![image](https://user-images.githubusercontent.com/63607175/204965172-bc109041-a781-4fd9-8a88-7735695fc5e0.png)

Hi @alexdrenea - works fine on my instance, could be a specific use case where this might fail. Additional Repro steps would help please.

santhoshb-msft:
@alexdrenea as per my above comments, closing this for now, please let me know if can repro this, we will reopen and fix.

Issue #350, opened 2022-11-07
Customer Portal Change Plan / Quantity functionality


alexdrenea:
**Describe the bug**
In the Customer Portal, we are disabling the Change Plan / Quantity functionality based on the wrong application setting:

https://github.com/Azure/Commercial-Marketplace-SaaS-Accelerator/blob/1d2805fd489595566fe6ab52b279947ea043d3f4/src/CustomerSite/Views/Home/Subscriptions.cshtml#L64

If I understand the settings correctly, we should be looking at "AcceptSubscriptionUpdates" and not "IsAutomaticProvisioningSupported" to show/hide the Change Plan / Quantity options.

dstarr:
Closing. Fixed in PR #399

PR #399, opened 2022-12-12
Refresh Plan Name

### LINKED ISSUE #353: Wrong SubscriptionStatus in Subscription Details page

alexdrenea:
### Logic:

- when changing plan, ensure we update the database with the new plan for the offer (and log)
- when fetching subscription list, ensure we refresh plan name, quantity or status and log

### UI:
- change plan / change quantity are long running operations (30s-60s). Disable button and show spinner on page while operation is in progress for both change plan and change quantity modals both in admin portal and customer portal
 
image

Addresses #345, #350, #219
Also fixes #353 , #354

### DIFF
--- a/src/AdminSite/Controllers/HomeController.cs
+++ b/src/AdminSite/Controllers/HomeController.cs
@@ -696,7 +696,6 @@ namespace Microsoft.Marketplace.Saas.Web.Controllers
                     {
                         //initiate change plan
                         var currentUserId = this.userService.GetUserIdFromEmailAddress(this.CurrentUserEmailAddress);
-                        var currentSubscription = this.subscriptionService.GetSubscriptionsBySubscriptionId(subscriptionDetail.Id);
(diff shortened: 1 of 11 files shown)


dstarr (changes requested):
Tested as far as the base behavior described in #345. The issue doesn't seem to be resolved. Ran a test in which
1. I insttalled the PR
2. Changed Plan via Azure portal and again through the admin site.
3. While the plan did change as per the Azure portal, no change occurred in the DB
4. The app logs show the webhook fired and the plan change was made, but the change was not made in the Subscriptions table of the DB

Also tested for #350 and the Change Plan option did not show up pn the landing page subscriptions for a subscription that definately has multiple plans.

dstarr (approved):

PR #411, opened 2022-12-14
Update casing on location variable

code4clouds:
Update casing on location variable to match the declaration casing.


santhoshb-msft (approved):
Tested deployed with West US and East US. Deployment success and applications working deployed to the locations and working as expected.

PR #415, opened 2022-12-19
DS refactoring offers - Stage 1

### LINKED ISSUE #375: Repository cleanup
**Problem:** Our code is using the classic Controller-Service-Repository pattern + build in DI offered by the framework . But it is not utilized correctly as both repositories and services are injected in controllers and used. For DI, services should be registered in the DI container and referenced in controllers rather than created on demand in controller ctors

**Solution:** Refactor code to make use of DI container. Refactor controllers to only reference Services and abstract away the Repository.

**Benefits:** Stability, Maintainability, Testability.

dstarr:
- Removed IOfferRepository from controller
- Added SaaSKitContext to controller to inti OfferService
- OfferService now provides all Offers
- OfferService has test contructor with IRepositories to facilitate mocks or stubs.
- OfferController no longer references IOfferAttributes repo.

Adddresses but not close #375


src/DataAccess/Migrations/20221118045814_Baseline_v2.cs
```diff
@@ -119,7 +119,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
                 });
 
             migrationBuilder.CreateTable(
-                name: "Offers",
+                name: "Offer",
```
alexdrenea, line 122: This will cause a problem for migrations. Unfortunately, we are stuck with table names as they were and we cannot easily change them and keep compatibility with older deployments (which is something that we want to maintain).

As a general rule, we should never touch existing migration files. 

I agree though that naming the class "Offers" is not ideal and support the name change to "Offer". However, we need to keep the table name as "Offers".

src/DataAccess/Services/OffersRepository.cs
```diff
@@ -26,55 +26,56 @@ public class OffersRepository : IOffersRepository
...
+        return this.dbContext.Offers.FirstOrDefault(s => s.Id == id);
     }
 
     /// <summary>
     /// Adds the specified offer details.
     /// </summary>
     /// <param name="offerDetails">The offer details.</param>
     /// <returns> Offer Id.</returns>
-    public Guid Add(Offers offerDetails)
+    public Guid Add(Offer offerDetails)
```
alexdrenea, line 61: Not sure if this is in scope for your current change, but though I'd put the comment in. 

What are your thoughts on the repositories' Add method to return _the object_ , or <T>, instead of int or Guid? This is something that we would eventually change everywhere but this could be a good start. 

Thoughts?

src/AdminSite/Controllers/OffersController.cs
```diff
@@ -60,16 +54,33 @@ public OffersController(IOffersRepository offersRepository, IApplicationConfigRe
...
     public IActionResult Index()
     {
-        this.logger.LogInformation("Offers Controller / Index");
+        this.logger.LogInformation("Offer Controller / Index");
         try
         {
-            List<OffersModel> getAllOffersData = new List<OffersModel>();
             this.TempData["ShowWelcomeScreen"] = "True";
+
+            OffersListViewModel viewModel = new OffersListViewModel()
```
alexdrenea, line 62: Style question... could we use LINQ to create our objects rather than initialize them ahead of time ?

```csharp
List<OfferModel> offers = this.offersService.GetAllOffers();
var lineItems = offers.Select(offer=> new OffersListViewModel.OfferListItem()
                {
                    OfferName = offer.OfferName,
                    OfferId = offer.OfferID,
                    OfferGuid = offer.OfferGuid.GetValueOrDefault()
                }).ToList();

var viewModel = new OffersListViewModel
{
    LineItems = lineItems
};

src/DataAccess/Services/OffersRepository.cs
dstarr: I like it. In this case I was trying go lightly on the repo changes. Believe it or not :(

src/AdminSite/Controllers/OffersController.cs
dstarr: In these cases, I am keeping entity and View Model mapping manual for now. I actually want to introduce AutoMapper to take care of these situations in the future.

dstarr:
I am closing this pull request to go back and ensure no changes to migration files.

Issue #379, opened 2022-12-04
Clean up Admin site controllers


dstarr:
Visit the MVC controller classes looking for 

- opportunities to split them into smaller controllers
- reducing the number of injected items
- removing repositories as injected items, moving to a repo factory model
- Using [FromService] where appropriate
- Adding unit tests

PR #417, opened 2022-12-20
DS offers repo refactoring - Part 1

dstarr:
Addresses, but does not close, #379 #375 

This is a very simple change and the first of several which will be required on the OffersController, but I wanted to send this in as it has a new test project and does remove at least one repo from the controller.

- Added very simple unit test project to test the offers service
- Added new contructior to IOffersRepository to take SaaSKitContext
- Put the IOffersRepo under the OfferServices class
- Removed IOffersRepo dependency from the OfferController class
- Refactored controller
- Deleted all commented out code in the OffersDetail view
- Fully tested the landing page
- Made no changes to the DataAccess project


src/AdminSite/Startup.cs
```diff
@@ -139,6 +139,8 @@ public void ConfigureServices(IServiceCollection services)
         services.Configure<CookieTempDataProviderOptions>(options => {
             options.Cookie.IsEssential = true;
         });
+
```
santhoshb-msft, line 142: Can we add this to the InitializeRepositoryServices method, @dstarr ?

dstarr: I don't see that as a repository. InitializeRepositoryServices is a private method that just wraps registering repos with the DI system.

santhoshb-msft: I am thinking if its reading like InitializeRepositorys and InitializeServices ?and having them at one place would help with whats used for DI at one place.

dstarr: I guess I don't understand as `Startup.ConfigureServices` is the containing function for registering services. The other methods you are referring to are private methods provided for clarity and brevity.

I may be misunderstanding, in which case we should just have a brief conversation.

santhoshb-msft (approved):
Looks good, @dstarr . Approved.

PR #418, opened 2022-12-21
Refactor Admin - ApplicationConfigController

alexdrenea:
Added new ActionFilters
- Error Handler filter to inject in controllers as needed
- Request Log Filter to inject in controllers
- Register as services in both customer and admin portals
- Add logging providers for debugging

Refactor ApplicationConfigController
- Implement request logging filter
- remove redundant log statements and simplify controller methods
- Get Service from DI instead of building it in the controller ctor
- fix PartialViews to only show Partial View Content (div and not body)
- update error messages to use swal instead of alert
- introduce Error middleware (Cannot apply at top level of controller due to mix of controller methods (some returning a View, some PartialView and some doing redirects). Errors are handled in the cshtml for those)
- fixes a bug where the try/catch in methods returning PartialView (ApplicationConfigDetails, EmailTemplateDetails) displays error page in modal (looks bad).

### DIFF
--- a/src/AdminSite/Controllers/ApplicationConfigController.cs
+++ b/src/AdminSite/Controllers/ApplicationConfigController.cs
@@ -18,6 +18,7 @@ namespace Marketplace.SaaS.Accelerator.AdminSite.Controllers;
 /// </summary>
 /// <seealso cref="BaseController" />
 [ServiceFilter(typeof(KnownUserAttribute))]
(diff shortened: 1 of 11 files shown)


dstarr (approved):
Tested all configuration setting page functionality and Landing Page just to be sure.

Love the central logging. Will use it for my PR.

PR #423, opened 2023-01-17
fix email logo

msalemcode:
Replace the email log with updating live link


alexdrenea:
Looks good to me! 
But wanted to make sure it's understood this will only work for new deployments of the accelerator. 
For anyone who's deployed it before this change, they would need to go and manually make the changes in the templates (though the db or admin UI)

msalemcode:
> 

@alexdrenea This is correct, it is will be for future deployments. Existing deployment will need manual fix

code4clouds (commented):
Why are these files not local on the src report?

code4clouds (changes requested):
Please change the logo to be the logo served by the landing page (https://<prefix>-portal.azurewebsites.net/contoso-sales.png).

code4clouds (approved):

PR #424, opened 2023-01-20
Use existing resources like KeyVault

msalemcode:
Enhancing the existing script to handle existing resources like keyvault, webapp, sql server ..etc

Also the script will allow user to pass keyvault name instead of autogenerated keyvault.

### DIFF
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -269,6 +269,10 @@ if(!($KeyVaultName))
 $KeyVault=$WebAppNamePrefix+"-kv"
 $KeyVault=$KeyVault -replace '_',''
 }
+else
+{
+  $KeyVault=$KeyVaultName
+}
 #keep the space at the end of the string - bug in az cli running on windows powershell truncates last char https://github.com/Azure/azure-cli/issues/10066
 $ADApplicationSecretKeyVault="@Microsoft.KeyVault(VaultName=$KeyVault;SecretName=ADApplicationSecret) "
 $DefaultConnectionKeyVault="@Microsoft.KeyVault(VaultName=$KeyVault;SecretName=DefaultConnection) "
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -296,7 +296,10 @@ else
 	Write-host "      ➡️ Rest User Password"	
 	az sql server update --name $SQLServerName  --resource-group $ResourceGroupForDeployment --admin-password $SQLAdminLoginPassword
 	$SQLAdminLogin=az sql server show --name $SQLServerName  --resource-group $ResourceGroupForDeployment --query administratorLogin
+	$SQLAdminLogin=$SQLAdminLogin-replace '\"',''
 }
+
+
 
 Write-host "      ➡️ Add SQL Server Firewall rules"
 az sql server firewall-rule create --resource-group $ResourceGroupForDeployment --server $SQLServerName -n AllowAzureIP --start-ip-address "0.0.0.0" --end-ip-address "0.0.0.0" --output $azCliOutput
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -282,7 +282,7 @@ az group create --location $Location --name $ResourceGroupForDeployment --output
 
 Write-host "   🔵 SQL Server"
 Write-host "      ➡️ Check if Sql Server Exists"
-if(-Not (az sql server show --name sass-msalem20231sql --resource-group $ResourceGroupForDeployment 2>null))
+if(-Not (az sql server show --name $SQLServerName  --resource-group $ResourceGroupForDeployment 2>null))
 {
  Write-host "      ➡️ Create Sql Server"	
  az sql server create --name $SQLServerName --resource-group $ResourceGroupForDeployment --location $Location --admin-user $SQLAdminLogin --admin-password $SQLAdminLoginPassword --output $azCliOutput
@@ -325,7 +325,7 @@ if(-Not (az appservice plan show -g $ResourceGroupForDeployment -n $WebAppNameSe
 
 Write-host "   🔵 Admin Portal WebApp"
 Write-host "      ➡️ Check if Web App exists"
-if(-Not (aaz webapp show -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNameAdmin 2>null))
+if(-Not (az webapp show -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNameAdmin 2>null))
 {
   Write-host "      ➡️ Create Web App"
   az webapp create -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNameAdmin  --runtime dotnet:6 --output $azCliOutput
@@ -341,7 +341,7 @@ az webapp config appsettings set -g $ResourceGroupForDeployment  -n $WebAppNameA
 az webapp config set -g $ResourceGroupForDeployment -n $WebAppNameAdmin --always-on true  --output $azCliOutput
 
 Write-host "   🔵 Customer Portal WebApp"
-if(-Not (aaz webapp show -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNamePortal 2>null))
+if(-Not (az webapp show -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNamePortal 2>null))
 {
   Write-host "      ➡️ Create Web App"
   az webapp create -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNamePortal --runtime dotnet:6 --output $azCliOutput
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -277,9 +277,12 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User 
 
 
 Write-host "   🔵 Resource Group"
+Write-host "      ➡️ Check Resource Group if exists"
+if(-Not (az group show -n $ResourceGroupForDeployment 2>null))
+{
 Write-host "      ➡️ Create Resource Group"
 az group create --location $Location --name $ResourceGroupForDeployment --output $azCliOutput
-
+}
 Write-host "   🔵 SQL Server"
 Write-host "      ➡️ Check if Sql Server Exists"
 if(-Not (az sql server show --name $SQLServerName  --resource-group $ResourceGroupForDeployment 2>null))
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -24,10 +24,12 @@ Param(
    [string][Parameter()]$SQLServerName, # Name of the database server (without database.windows.net)
    [string][Parameter()]$SQLAdminLogin, # SQL Admin login
    [string][Parameter()]$SQLAdminLoginPassword, # SQL Admin password  
+   [string][Parameter()]$KeyVaultName, # KeyVault Name
    [string][Parameter()]$LogoURLpng,  # URL for Publisher .png logo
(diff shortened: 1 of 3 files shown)


deployment/Deploy.ps1
```diff
@@ -262,8 +264,11 @@ Write-host "☁ Deploy Azure Resources"
 $WebAppNameService=$WebAppNamePrefix+"-asp"
 $WebAppNameAdmin=$WebAppNamePrefix+"-admin"
 $WebAppNamePortal=$WebAppNamePrefix+"-portal"
+if(!($KeyVaultName))
+{
 $KeyVault=$WebAppNamePrefix+"-kv"
 $KeyVault=$KeyVault -replace '_',''
+}
```
code4clouds, line 271: Are you missing the else { $KeyVault = $KeyVaultName } somehere?  I see no place where the new $KeyVaultName is assigned to the $KeyVault

deployment/Deploy.ps1
```diff
@@ -272,36 +277,74 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User
...
 if ($env:ACC_CLOUD -eq $null){
     Write-host "      ➡️ Running in local environment - Add current IP to firewall"
 	$publicIp = (Invoke-WebRequest -uri "http://ifconfig.me/ip").Content
     az sql server firewall-rule create --resource-group $ResourceGroupForDeployment --server $SQLServerName -n AllowIP --start-ip-address "$publicIp" --end-ip-address "$publicIp" --output $azCliOutput
 }
-Write-host "      ➡️ Create SQL DB"
-az sql db create --resource-group $ResourceGroupForDeployment --server $SQLServerName --name $SQLDatabaseName  --edition Standard  --capacity 10 --zone-redundant false --output $azCliOutput
+
+Write-host "      ➡️ Check if SQL DB exists"
+if(-Not (az sql db show --server $SQLServerName --name $SQLDatabaseName --resource-group $ResourceGroupForDeployment 2>null))
```
code4clouds, line 314: Hard to read... would you extract this out?

deployment/Deploy.ps1
```diff
@@ -312,8 +355,12 @@ az webapp config appsettings set -g $ResourceGroupForDeployment  -n $WebAppNameA
 az webapp config set -g $ResourceGroupForDeployment -n $WebAppNameAdmin --always-on true  --output $azCliOutput
 
 Write-host "   🔵 Customer Portal WebApp"
-Write-host "      ➡️ Create Web App"
-az webapp create -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNamePortal --runtime dotnet:6 --output $azCliOutput
+if(-Not (az webapp show -g $ResourceGroupForDeployment -n $WebAppNamePortal 2>null))
```
code4clouds, line 358: Hard to read code... extract this to a variable to be compared...

docs/Installation-Instructions.md
```diff
@@ -52,7 +52,7 @@ git clone https://github.com/Azure/Commercial-Marketplace-SaaS-Accelerator.git -
 cd ./Commercial-Marketplace-SaaS-Accelerator/deployment; `
 .\Upgrade.ps1 `
  -WebAppNamePrefix "marketplace-SOME-UNIQUE-STRING" `
- -ResourceGroupForDeployment "marketplace-SOME-UNIQUE-STRING"" `
+ -ResourceGroupForDeployment "marketplace-SOME-UNIQUE-STRING"
```
code4clouds, line 55: leave the back ticks there are easy on the users when they are copy and pasted the additional paremeters

docs/Installation-Instructions.md
```diff
@@ -18,7 +18,7 @@ cd ./Commercial-Marketplace-SaaS-Accelerator/deployment; `
  -WebAppNamePrefix "marketplace-SOME-UNIQUE-STRING" `
  -ResourceGroupForDeployment "marketplace-SOME-UNIQUE-STRING" `
  -PublisherAdminUsers "user@email.com" `
- -Location "East US" `
+ -Location "East US"
```
code4clouds, line 21: leave the back ticks there are easy on the users when they are copy and pasted the additional paremeters

code4clouds (changes requested):

deployment/Deploy.ps1
```diff
@@ -272,36 +277,74 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User
 
 
 Write-host "   🔵 Resource Group"
+Write-host "      ➡️ Check Resource Group if exists"
+if(-Not (az group show -n $ResourceGroupForDeployment 2>null))
```
alexdrenea, line 281: There's really no reason to do this...
the az group create command already does that for you behind the scenes. If the RG exists, nothing happens, it if doesn't it gets created. 

This logic just adds unnecessary complexity to our script.

deployment/Deploy.ps1
```diff
@@ -272,36 +277,74 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User
...
+{
+	Write-host "      ➡️ Create SQL DB"	
+  az sql db create --resource-group $ResourceGroupForDeployment --server $SQLServerName --name $SQLDatabaseName  --edition Standard  --capacity 10 --zone-redundant false --output $azCliOutput
+}
 
 Write-host "   🔵 KeyVault"
-Write-host "      ➡️ Create KeyVault"
-az keyvault create --name $KeyVault --resource-group $ResourceGroupForDeployment --output $azCliOutput
+Write-host "      ➡️ Check KeyVault if exists"
+if(-Not (az keyvault show --name $KeyVault --resource-group $ResourceGroupForDeployment 2>null))
```
alexdrenea, line 322: Same as for ResourceGroup.
`az keyvault` command will do this for you.

deployment/Deploy.ps1
```diff
@@ -272,36 +277,74 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User
...
+
 Write-host "      ➡️ Add Secrets"
 az keyvault secret set --vault-name $KeyVault  --name ADApplicationSecret --value $ADApplicationSecret --output $azCliOutput
 az keyvault secret set --vault-name $KeyVault  --name DefaultConnection --value $Connection --output $azCliOutput
 
 Write-host "   🔵 App Service Plan"
-Write-host "      ➡️ Create App Service Plan"
-az appservice plan create -g $ResourceGroupForDeployment -n $WebAppNameService --sku B1 --output $azCliOutput
+Write-host "      ➡️ Check App Service Plan if exists"
+if(-Not (az appservice plan show -g $ResourceGroupForDeployment -n $WebAppNameService 2>null))
```
alexdrenea, line 334: Same as for ResourceGroup.
`az appservice` command will do this for you.

deployment/Deploy.ps1
```diff
@@ -272,36 +277,74 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User
...
+{
+  Write-host "      ➡️ Create App Service Plan"
+  az appservice plan create -g $ResourceGroupForDeployment -n $WebAppNameService --sku B1 --output $azCliOutput
+}
 
 Write-host "   🔵 Admin Portal WebApp"
-Write-host "      ➡️ Create Web App"
-az webapp create -g $ResourceGroupForDeployment -p $WebAppNameService -n $WebAppNameAdmin  --runtime dotnet:6 --output $azCliOutput
+Write-host "      ➡️ Check if Web App exists"
+if(-Not (az webapp show -g $ResourceGroupForDeployment -n $WebAppNameAdmin 2>null))
```
alexdrenea, line 342: Same as for ResourceGroup.
`az webapp` command will do this for you.

deployment/Deploy.ps1
alexdrenea, line 358: Same as for ResourceGroup.
`az webapp` command will do this for you.

deployment/Deploy.ps1
```diff
@@ -272,36 +277,74 @@ $Connection="Data Source=tcp:"+$ServerUri+",1433;Initial Catalog=AMPSaaSDB;User
...
+{
 Write-host "      ➡️ Create Resource Group"
 az group create --location $Location --name $ResourceGroupForDeployment --output $azCliOutput
-
+}
 Write-host "   🔵 SQL Server"
-Write-host "      ➡️ Create Sql Server"
-az sql server create --name $SQLServerName --resource-group $ResourceGroupForDeployment --location $Location --admin-user $SQLAdminLogin --admin-password $SQLAdminLoginPassword --output $azCliOutput
+Write-host "      ➡️ Check if Sql Server Exists"
+if(-Not (az sql server show --name $SQLServerName  --resource-group $ResourceGroupForDeployment 2>null))
```
alexdrenea, line 288: SQL is the only exception where we could make the improvements of checking existence and updates. 
But the problem isn't with creating the sql server (the az commands will automatically do that), the problem is updating the password in the keyvault.

If the server already exists, we need to provide the password, otherwise we should fail the script.

I think resetting the password every time we deploy might cause issues. What is a user deployed the accelerator in an existing server with their own password. We are messing around with resources and passwords and we can't guarantee that the server is not being used in other places

deployment/Deploy.ps1
alexdrenea: I agree with Julio! 
The way you did the checks is a bit hard to read.
Using the "show" command usually throws an error if resource is not found which is why you needed to use the standard output redirection.

I would recommend using the "list" command for easier processing:

```powershell
$sqlName = az sql server list --query "[?name=='$SQLServerName'].{name:name}" --output tsv
if ($sqlServer -ne $null){
 //Exists
} else {
 //Does not exist
}
```

Do note that some services expose a specific "exists" command that you can use (resourceGroup, storage)

```powershell
$rgExists = az group exists --name $resourceGroup 
if ($rgExists -eq 'true'){
 //Exists
} else {
 //Does not exist
} 
```

For others you can query by using the "list" command.

deployment/Deploy.ps1
msalemcode: This issue is fix by adding else clause

deployment/Deploy.ps1
msalemcode: This Show command structure in a way it will not throw error

docs/Installation-Instructions.md
msalemcode: I removed it based on the ISV feedback they get confused when run the script and it stop waiting for another enter because of the extra ticks

docs/Installation-Instructions.md
msalemcode: see above comment

deployment/Deploy.ps1
msalemcode: I tried to add the same logic for all resources to avoid any unwanted errors that can confuse customer with messages they do not understand.

deployment/Deploy.ps1
msalemcode: see above comment

deployment/Deploy.ps1
msalemcode: see above comment

deployment/Deploy.ps1
msalemcode: see above comment

deployment/Deploy.ps1
msalemcode: Resetting password will cover the cases where use did not use a custom password and used the autogenerated one. the script will auto generate new one and update KV. this case will not mess around with the resources.

IF user would like to use specific password and user then they can pass the SQL User and password as part of the script parameters

Also, the main use case to Re-Running script against same resources will be when customer run the script for first time and it fails for some reason. they will have ability to re-run it again.

PR #426, opened 2023-01-24
added validator with regest for sql password

code4clouds:


santhoshb-msft (approved):
@code4clouds - 
Tested with multiple combinations. works as expected. Approved.

image
image

PR #428, opened 2023-01-26
add the keyvault paramter

msalemcode:
Adding Key Vault as parameter so user can provide name for the key vault to create or use existing one

### DIFF
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -26,6 +26,8 @@ Param(
    [string][Parameter()]$SQLAdminLoginPassword, # SQL Admin password  
    [string][Parameter()]$LogoURLpng,  # URL for Publisher .png logo
    [string][Parameter()]$LogoURLico,  # URL for Publisher .ico logo
+   [string][Parameter()]$LogoURLico,  # URL for Publisher .ico logo
+   [string][Parameter()]$KeyVault, # Name of KeyVault
    [switch][Parameter()]$MeteredSchedulerSupport, # set to true to enable Metered Support
    [switch][Parameter()]$Quiet #if set, only show error / warning output from script commands
 )
@@ -74,6 +76,14 @@ if($WebAppNamePrefix.Length -gt 21) {
     Exit
 }
 
+if ($string -contains "find") { 
+	# Do something
+  }
+
+if($KeyVault -contains "_") {
+    Throw "🛑 KeyVault name can not contain hyphens"
+    Exit
+}
 #endregion 
 
 Write-Host "Starting SaaS Accelerator Deployment..."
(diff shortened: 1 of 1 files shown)


code4clouds (changes requested):

deployment/Deploy.ps1
```diff
@@ -262,8 +268,12 @@ Write-host "☁ Deploy Azure Resources"
 $WebAppNameService=$WebAppNamePrefix+"-asp"
 $WebAppNameAdmin=$WebAppNamePrefix+"-admin"
 $WebAppNamePortal=$WebAppNamePrefix+"-portal"
-$KeyVault=$WebAppNamePrefix+"-kv"
-$KeyVault=$KeyVault -replace '_',''
+if(! ($KeyVault))
```
alexdrenea, line 271: Could we add this to the top of the file (lines 42-51) and use the same pattern for defaults that we use in there? 
Something like...
if ($KeyVault -eq "") {
    $KeyVault = $WebAppNamePrefix + "-kv"
}

deployment/Deploy.ps1
```diff
@@ -74,6 +75,11 @@ if($WebAppNamePrefix.Length -gt 21) {
     Exit
 }
 
+
+if($KeyVault -contains "_") {
```
alexdrenea, line 79: I think if we are validating characters, we should do a more complete validation (KVs can only contain alphanumeric and dashes). 

we could use a regex expression to either check (`$KeyVault -match "[^a-z0-9-]"`), or automatically correct the value by removing invalid characters (`$KeyVault = $KeyVault -replace "[^a-z0-9-]", "" `)

code4clouds: It's better that it fails for the customers to know the value enter will not be applied.

code4clouds: This is the right pattern
` $KeyVault -match "^[a-z0-9-]+$"`

deployment/Deploy.ps1
```diff
@@ -74,6 +75,11 @@ if($WebAppNamePrefix.Length -gt 21) {
...
+    Throw "🛑 KeyVault name can not contain hyphens"
```
code4clouds, line 80: The code checks for underscore _ and the message says hyphen -

code4clouds: From the docs hyphens - are good, underscores _ are not.

code4clouds (changes requested):
Just three small changes and its all good.

code4clouds (approved):

deployment/Deploy.ps1
```diff
@@ -262,8 +273,12 @@ Write-host "☁ Deploy Azure Resources"
 $WebAppNameService=$WebAppNamePrefix+"-asp"
 $WebAppNameAdmin=$WebAppNamePrefix+"-admin"
 $WebAppNamePortal=$WebAppNamePrefix+"-portal"
-$KeyVault=$WebAppNamePrefix+"-kv"
-$KeyVault=$KeyVault -replace '_',''
+if(! ($KeyVault))
```
alexdrenea, line 276: I don't this this check is needed anymore because we set the KeyVault value in line 58-61 and we check that the name is valid with the Regex expression in lines 84-87 so there is no chance that the keyvault is empty or invalid at this point.

Otherwise looks good!

alexdrenea (approved):

santhoshb-msft:
@msalemcode default installation script failing
image

deployment/Deploy.ps1
```diff
@@ -74,6 +80,11 @@ if($WebAppNamePrefix.Length -gt 21) {
     Exit
 }
 
+
+if($KeyVault -match "^[a-z0-9-]+$") {
```
alexdrenea, line 84: this condition is reversed!

If the value matches the regex, that means it's a **valid** value so we should **not** throw.

Please reverse the condition. thank you @santhoshb-msft

msalemcode: Condition is fixed

santhoshb-msft (approved):
deployed the changed script multiple times with and without specifying the kv. works good. Approved.

PR #433, opened 2023-01-31
Passed fulfilment base url to Marketplace SaaS client

scottperham:
There is a configuration setting to allow the base url for the sass fulfilment api to be changed from the default, but the value was never used.

### DIFF
--- a/src/AdminSite/Startup.cs
+++ b/src/AdminSite/Startup.cs
@@ -111,7 +111,10 @@ public class Startup
             .AddScoped<RequestLoggerActionFilter>()
         ;
 
-        var fulfillmentBaseApi = new Uri(config.FulFillmentAPIBaseURL ?? "https://marketplaceapi.microsoft.com/api");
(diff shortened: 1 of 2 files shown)


mormond:
@alexdrenea @santhoshb-msft This change would be really helpful for us to enable partners to test with the SaaS Accelerator. Just to add an upvote :)

santhoshb-msft:
Hi @mormond yes this will go in this week. I'm testing on my side. 
@alexdrenea look forward to your comments.

alexdrenea (approved):
Changes look good to me!

santhoshb-msft:
Thanks @alexdrenea 
@scottperham/ @mormond  - the change looks good, just one thing, can we handle the empty string case please?

image

scottperham:
Updated to ensure the configured URI is a valid absolute URL

santhoshb-msft:
Thank you pushing the new changes. @scottperham. Reviewing them now.

santhoshb-msft (approved):
Thank you for this change, @scottperham .
Tested and Approved

PR #454, opened 2023-03-17
Added in missing space 

tiffwchen:
Added in missing space for message shown on customer portal subscriptions page when there are no subscriptions.


santhoshb-msft (approved):
thank you, @tiffwchen approved

Issue #462, opened 2023-03-30
Filter the schedule frequency options based on the App Config setting


msalemcode:
Currently App Config require users to enable specific schedule in order for the scheduler engine to fire it. However user still can add scheduler event for disabled option.

The application should show the enabled schedule frequency so user can only schedule the option enabled for engine

PR #463, opened 2023-03-31
add filter to scheduler frequency

msalemcode:
This PR to address bug #462 

The change will enhance "Add New Scheduler" action to show only enabled run schedule frequency.


santhoshb-msft (commented):
@msalemcode couple of things
1. Does this mean when I disable a frequency after the scheduler event is created, it will still run?
2. Can we add a note that if they dont see anything in the dropdown, they need to enable in the app settings.

src/AdminSite/Views/Scheduler/NewScheduler.cshtml
```diff
@@ -92,6 +92,12 @@
                     <tr>
                         <th>Frequency</th>
                        <th>@Html.DropDownListFor(s => s.SelectedSchedulerFrequency, Model.SchedulerFrequencyList, "-- Select Frequency --", new { @class = "form-control", required = "required" })</th>
+                        @if (Model.SchedulerFrequencyList.Count() == 0 )
+                        {
+                            <th>
+                                <span style="color:red">Please enable the frequecies from App Config!</span>
```
code4clouds, line 98: The message needs more information.   As in "Please enable frequencies from the App Config, to allow the system to .... " 

Also, no need for using exclamation mark...

msalemcode: Please review the latest update

code4clouds (approved):

PR #465, opened 2023-04-04
Added in-memory database for local testing

scottperham:
Added EF InMemory database provider for local testing without SQL dependency.


santhoshb-msft (commented):
@scottperham this is great but just a couple of thoughts.
1. can we add a warning on the pages , if the in-memory db is used please
2. is there a way we can only have this enabled in localhost and/or debug mode

Trying to avoid the risk of someone accidentally using this and realizing later. Please let me know what you think

scottperham:
@santhoshb-msft I have added the following message to the top of the pages for both the customer site and admin site when the in-memory database provider is being used...

image

santhoshb-msft:
Hi @scottperham could you please help resolve the conflicts.

santhoshb-msft:
@scottperham I think it would help if this is driven by the appsetting. Developermode appsetting and no connectionstring then enable the inmemory db please?

santhoshb-msft:
@scottperham please let me know if we can do that above and resolve some conflicts.

santhoshb-msft:
@scottperham please let me know if you can fix as per comments

santhoshb-msft:
Closing this as no response. Please feel to reopen once the conflicts are resolved with the above feedback.

Issue #435, opened 2023-01-31
Meter scheduler page not showing dimension


code4clouds:
**Describe the bug**
The meter dimension dropdown does not show values.

**To Reproduce**
1. Select a subscription with meters
2. Select "manage usage"
3. without selecting a meter, click "add schedule usage"
4. populate the form but notice the dimension dropdown is empty.

**Expected behavior**
The schedule meter dimension dropdown shows all the meters available.

**Screenshots**
![image](https://user-images.githubusercontent.com/29647050/215902671-dbcdfb19-3b11-4c1f-9755-5ab6562f2933.png)

![image](https://user-images.githubusercontent.com/29647050/215902783-56577c0a-9373-4b61-b9a6-d5f50384bba1.png)

santhoshb-msft:
Currently when 
1. When we go to new scheduler page, is empty and everything populates as we select
2. when we click the Add Scheduler button when Dimension is selected, the New Scheduler page pre-populates only that dimension, 
3. when we click the Add Scheduler button and dimension is selected, there was no dimension available in the dropdown. @tiffwchen fixed to load the dimensions

PR #468, opened 2023-04-07
Fix For Meter Scheduler Bug #435

tiffwchen:
Fixed bug with meter scheduler where the dimension dropdown is empty if the dimension is not selected for manage usage to add schedule usage workflow.


santhoshb-msft (approved):
Thank you @tiffwchen

PR #479, opened 2023-04-21
[WIP] Standalone script for registering Web Aps

jeernest:


deployment/Deploy.ps1
```diff
@@ -101,21 +103,41 @@ if(!($TenantID)) {
...
 }
 else {
-    Write-Host "🔑 Tenant provided: $TenantID"
+    Write-Host "🔑 Tenant provided for Web App Registrations: $TenantID"
 }
 
 #Get Azure Subscription if not set as argument
 if(!($AzureSubscriptionID)) {    
-    Get-AzSubscription -TenantId $TenantID | Format-Table
+    Get-AzSubscription | Format-Table
```
jeernest, line 111: The tenant can be different, for example in case of development there can be some testing AAD out of company.

santhoshb-msft:
Hi @jeernest , we really appreciate you taking time on making this changes. We will review this internally and get back to you. We might not want to split the files and rather keep the logic in one place. But I do agree on some updates we could potentially make. I will keep you posted. Once again sincerely appreciate all your efforts here.

jeernest:
> Hi @jeernest , we really appreciate you taking time on making this changes. We will review this internally and get back to you. We might not want to split the files and rather keep the logic in one place. But I do agree on some updates we could potentially make. I will keep you posted. Once again sincerely appreciate all your efforts here.

Actually, the Deploy script still does the same logic and should be backward compatible. It just adds the possibility to specify a tenant different from the subscription. If you look at the current logic, the tenantID is required in 2 places:

* selecting a subscription - and I don't think this is neccessary, I'm adding an ability to choose whatever subscription is available.  Even developer's monthly subscription with $200 free credit, which might not be part of the same AAD. I don't think that it's a big deal if users will see little bit more subscriptions.
* creating the web registration - I propose to keep this variable for that purpose, because here it is really important.

The reason why I created a standalone RegisterApps.ps1 is that it can be passed to DevOps more easily if developer doesn't have the right to create registrations.

I personally don't like big scripts since I think it's problematic for maintenance. But I understand that you have your reasons, in such case it should be possible to add -RegisterAppsOnly switch (or something similar), which stops just after registering web apps resulting in the same functionality provided by the RegisterApps.ps1.

If you agree with the logic, I'll do thorough testing based on scenarios recommended.

code4clouds:
@jeernest - We are still talking about the PR... The separation of files is the main theme.  Would it be possible to keep the registration on the same file as we had it if required?

code4clouds:
@alexdrenea Thoughts on separating the registration to separate .ps files?

deployment/Deploy.ps1
jeernest: it is common to have testing AAD with [onmicrosoft.com domain](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-create-new-tenant) for development.

deployment/Deploy.ps1
```diff
@@ -34,6 +34,7 @@ Param(
 # Make sure to install Az Module before running this script
 # Install-Module Az
 # Install-Module -Name AzureAD
+# Install-Module SqlServer
```
jeernest, line 37: there's Invoke-Sqlcmd cmdlet used for database deployment

jeernest:
Closing as there's no easy backward-compatible way how to rewrite the current script to allow only registering apps and because there's no meaningful response from owners.

PR #484, opened 2023-04-26

neelavarshad:
Changing character limit documented to 21


santhoshb-msft (approved):
looks good, @neelavarshad

PR #498, opened 2023-05-18
release 7.2.0

santhoshb-msft:
- updated install instructions to the new release for normal and advanced
- updated security fixes instructions to the top


santhoshb-msft (commented):
self-approving

PR #499, opened 2023-05-19
Added message on staying up with new releases

dstarr:
Just a documentation update encouraging people to stay current on new releases.


santhoshb-msft (approved):
looks good approved

PR #500, opened 2023-05-19
Update Deploy.ps1

neelavarshad:
Added numeric check for keyvault names

### DIFF
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -81,10 +81,11 @@ if($WebAppNamePrefix.Length -gt 21) {
 }
 
 
-if(!($KeyVault -match "^[a-z0-9-]+$")) {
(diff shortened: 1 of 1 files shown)


santhoshbomma9:
@neelavarshad - could you please add an issue to this PR?

neelavarshad:
Added Issue #512

santhoshb-msft:
@louisephilardeau is working on testing this

louiseph:
Hi @neelavarshad, the PR looks good to me, can confirm after testing that the regex no longer allows for KeyVault name to start with a number, and all other existing behavior still works as expected. One quick note: The change in regex also excludes the name from starting with a hyphen - assuming this is desired, would it make sense to call this out in the error message by specifying that the name can only start with a letter?

louiseph (approved):

PR #513, opened 2023-05-31
added aft token validation

santhoshb-msft:


santhoshb-msft:
Closing this to replace with #514

PR #514, opened 2023-05-31
added aft validation

santhoshb-msft:
added aft validation


alexdrenea (approved):

alexdrenea:
Reviewed this! 
The Attribute was added on all the relevant controllers and the front-end calls were updated to send the header in the odd case where the POST was not submitting the form

santhoshb-msft:
Thank a lot for your help, @alexdrenea

PR #533, opened 2023-07-22
Mitigated CodeQL logging security Alerts

msalemcode:
This PR to fix CodeQL logging security alerts.
1- Centralize logging to use a common library under services
2- Adding Encoding methods to encode user input to prevent sql injection
3- Prevent Exposure of private information

### DIFF
--- a/src/AdminSite/Controllers/ApplicationLogController.cs
+++ b/src/AdminSite/Controllers/ApplicationLogController.cs
@@ -28,7 +28,7 @@ public class ApplicationLogController : BaseController
     }
     public IActionResult Index()
     {
-        ProcessInformation("Application Log Controller / Index");
+        logger.ProcessInformation("Application Log Controller / Index");
         try
         {
             IEnumerable<ApplicationLog> getAllAppLogData = new List<ApplicationLog>();
--- a/src/AdminSite/Controllers/ApplicationLogController.cs
+++ b/src/AdminSite/Controllers/ApplicationLogController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Web;
 using Marketplace.SaaS.Accelerator.DataAccess.Contracts;
 using Marketplace.SaaS.Accelerator.DataAccess.Entities;
 using Marketplace.SaaS.Accelerator.Services.Services;
@@ -27,7 +28,7 @@ public class ApplicationLogController : BaseController
     }
     public IActionResult Index()
     {
-        this.logger.LogInformation("Application Log Controller / Index");
+        ProcessInformation("Application Log Controller / Index");
         try
         {
             IEnumerable<ApplicationLog> getAllAppLogData = new List<ApplicationLog>();
@@ -39,5 +40,11 @@ public class ApplicationLogController : BaseController
             this.logger.LogError($"Message:{ex.Message} :: {ex.InnerException}");
             return this.View("Error", ex);
         }
+    }
+
+    private void ProcessInformation(string message)
+    {
+        string log = HttpUtility.HtmlEncode(message);
+        this.logger.Info(log);
     }
 }
--- a/src/AdminSite/Controllers/ApplicationLogController.cs
+++ b/src/AdminSite/Controllers/ApplicationLogController.cs
@@ -22,7 +22,7 @@ public class ApplicationLogController : BaseController
     public ApplicationLogController(IApplicationLogRepository applicationLogRepository, SaaSClientLogger<ApplicationLogController> logger)
     {
         this.appLogRepository = applicationLogRepository;
-        this.logger = logger;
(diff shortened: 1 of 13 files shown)


src/AdminSite/Controllers/HomeController.cs
```diff
@@ -132,15 +132,15 @@ public class HomeController : BaseController
...
     {
         this.billingApiService = billingApiService;
         this.subscriptionRepo = subscriptionRepo;
         this.subscriptionLogRepository = subscriptionLogsRepo;
         this.planRepository = planRepository;
         this.subscriptionUsageLogsRepository = subscriptionUsageLogsRepository;
         this.dimensionsRepository = dimensionsRepository;
-        this.logger = logger;
+        this.tracker = logger;
```
dstarr: Why is this logger named 'tracker'?

src/AdminSite/Controllers/ApplicationLogController.cs
```diff
@@ -22,7 +22,7 @@ public class ApplicationLogController : BaseController
     public ApplicationLogController(IApplicationLogRepository applicationLogRepository, SaaSClientLogger<ApplicationLogController> logger)
     {
         this.appLogRepository = applicationLogRepository;
-        this.logger = logger;
+        this.logger = new SaaSClientLogger<ApplicationLogController>();
```
dstarr, line 26: Should this value be injected?

I think maybe: `this.logger = logger`

src/AdminSite/Startup.cs
```diff
@@ -207,7 +211,12 @@ private static void InitializeRepositoryServices(IServiceCollection services)
         services.AddScoped<ISchedulerFrequencyRepository, SchedulerFrequencyRepository>();
         services.AddScoped<IMeteredPlanSchedulerManagementRepository, MeteredPlanSchedulerManagementRepository>();
         services.AddScoped<ISchedulerManagerViewRepository, SchedulerManagerViewRepository>();
-      
-
+        services.AddScoped<SaaSClientLogger<HomeController>>();
```
dstarr, line 214: Maybe do an extract method refactoring to isloate the logger registrations? Just a style thing, not needed, but would make the containing method smaller and perhaps easier to read.

src/Services/Utilities/SaaSClientLogger.cs
```diff
@@ -113,11 +112,17 @@ public void Warn(string message, Exception ex)
...
+        this.logger.LogInformation(log);
     }
 
     public void LogError(string message)
     {
-        this.LogError(message);
+        this.logger.LogError(message);
+    }
+
+    public void ProcessInformation(string message)
```
dstarr: It's probably me because I haven't loaded the code into VS, but I can't find any references to this method anywhere. Curious why I would use it.

dstarr (commented):
Just a couple of questions.

src/AdminSite/Controllers/HomeController.cs
msalemcode: This is outdated code, the latest code does not have tracker

src/AdminSite/Controllers/ApplicationLogController.cs
msalemcode: correct this line is injection

src/Services/Utilities/SaaSClientLogger.cs
msalemcode: this is outdate class. the latest version does not have this class

santhoshb-msft:
@msalemcode there seems to be some conflicts, could you please resolve them?

src/AdminSite/Startup.cs
msalemcode: This methods was fixed in the new version as per our discussion

src/AdminSite/Controllers/KnownUsersController.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Web;
```
code4clouds, line 3: Did you add a code that requires the System.Web?

src/Services/Utilities/SaaSClientLogger.cs
```diff
@@ -12,27 +13,26 @@ namespace Marketplace.SaaS.Accelerator.Services.Utilities;
...
 /// </summary>
 /// <seealso cref="Microsoft.Marketplace.SaaS.SDK.Services.ILogger" />
-public class MeteringApiClientLogger : ILogger
+public class SaaSClientLogger<T> : ILogger
 {
     /// <summary>
     /// The logger.
     /// </summary>
-    private readonly ILogger<MeteredBillingApiService> logger;
+    private  ILogger<T> logger;
```
code4clouds, line 21: readonly made sense on the old code... If this is not going to be re-assigned later, it makes sense to have it.

code4clouds (approved):
Minor comments.

PR #537, opened 2023-08-08
Adding scheduler notification

msalemcode:
Enable Scheduler engine to send email notification to ISV for missing, successful and failure metered event.
Also, logging the engine events into App logging table


src/Services/Helpers/EmailHelper.cs
```diff
@@ -121,4 +121,64 @@ public EmailContentModel PrepareEmailContent(Guid subscriptionID, Guid planGuId,
...
+    /// <param name="schedulerTaskName">scheduler Task Name.</param>
+    /// <param name="responseJson">response Json.</param>
+    /// <param name="subscriptionStatus">The subscription status.</param>
+    /// <returns>
+    /// Email Content Model.
+    /// </returns>
+    /// <exception cref="Exception">Error while sending an email, please check the configuration.
+    /// or
+    /// Error while sending an email, please check the configuration.</exception>
+    public EmailContentModel PrepareMeteredEmailContent(string schedulerTaskName, String subscriptionName, string subscriptionStatus, string responseJson)
```
code4clouds, line 138: PrepareMeteredEmailContent and PrepareEmailContent are similar

code4clouds: It seems like we need a base class for setting the defaul emailcontent properties and then extend for the specific usage.

code4clouds (changes requested):
It seems like we need a base class for setting the default emailcontent properties and then extend for the specific usage on both PrepareMeteredEmailContent and PrepareEmailContent

src/Services/Helpers/EmailHelper.cs
msalemcode: cleaned up the code and it is optimized now

code4clouds (approved):

PR #543, opened 2023-08-29
Fixes az ADApplicationSecret hyphen parsing error

tpberntsen:
Fixes a bug where an ADApplicationSecret containing a starting hyphen is not parsed correctly, resulting in an "argument --value: expected one argument" error when executing the Deploy.ps1 script.


santhoshb-msft (approved):
tested by setting the -- and looks like the current command fails but we have not seen that happen why kv reset as it typically doesnt generate a secret value starting with --. But still a good one just in case, Thank you for your contribution, @tpberntsen

PR #546, opened 2023-08-31
updated test project dependency packages

santhoshb-msft:
Updated UI test projects to latest version


santhoshb-msft:
Merging as the change related to UI projects.

Issue #548, opened 2023-09-06
Billing Details on Admin Page


code4clouds:
**Is your feature request related to a problem? Please describe.**
There is no billing information on the subscription details

**Describe the solution you'd like**
I would like additional information about the subscription like plan term, start, end dates.

As an admin user that creates tenants for the SaaS solution
I would like to be able to see on the admin page the terms, start and end dates of the subscription,
So I don't have to customize this myself on the accelerator (out-the-box UX.

alexdrenea:
I have heard this from one partner as well!
If everyone is ok, happy to take this on.

santhoshb-msft:
This merge hence closing

PR #549, opened 2023-09-12
Added new properties to the Subscriptions table

alexdrenea:
- OfferId (Name) and Term details (Name, StartDate, EndDate)
- Updated subscriptions page both admin and customer to show Offer Name
- Updated subscriptions details page to show term details

Note!
For customers upgrading from an older version of accelerator, they should use the upgrade.ps1 script to upgrade the code and database, then run a "Fetch all Subscriptions" to populate the new fields.

Addresses #548 

Subscriptions page:
image

Subscription Details Page:
image


src/DataAccess/Migrations/20230912052848_SubscriptionDetails.cs
santhoshb-msft, line 1: @alexd would it be possible to name these files ending with _v714 to denote the future version please?

alexdrenea: done, renamed to _v740

santhoshb-msft (changes requested):
Thank you for making the migration and Ui changes. 
@alexdrenea could you please merge your branch from main latest.

santhoshb-msft (approved):
verified in both new and upgrade scenarios.

src/DataAccess/Migrations/20230912052848_SubscriptionDetails.cs
santhoshb-msft: thank you @alexdrenea

PR #554, opened 2023-09-21

code4clouds:
Added a sample with commas to illustrate multiple users and avoid ISVs using a single user as an admin.


santhoshb-msft (approved):
Usman and I tested this format and works good.

PR #577, opened 2023-10-18
New scheduler - minutes fix

jeernest:
**STR**
let's suppose that current time is 5:15pm

**current behavior**
1. create a new scheduler to run at 4:15pm - the form validation doesn't allow saving, because the time is in the past
2. create a new scheduler to run at 5:45pm - the form validation allows saving, because it is in the future, the StartDate is set to 6pm
3. create a new scheduler to run at 5:15pm - the form validation allows saving, because it is now, the StartDate is set to 5pm, which is in the past

**expected bahavior**
The form says that it rounds the minutes, so in case of current time, it should round the time to nearest hour in the future

### DIFF
--- a/src/DataAccess/Services/MeteredPlanSchedulerManagementRepository.cs
+++ b/src/DataAccess/Services/MeteredPlanSchedulerManagementRepository.cs
@@ -63,7 +63,7 @@ public class MeteredPlanSchedulerManagementRepository : IMeteredPlanSchedulerMan
         if (entity.StartDate.HasValue)
         {
             int minute = entity.StartDate.Value.Minute;
-            if (entity.StartDate.Value.Minute >= 30)
+            if (entity.StartDate.Value.Minute >= 30 || entity.StartDate.Value.ToUniversalTime() < DateTime.UtcNow)
             {
                 minute = 60 - entity.StartDate.Value.Minute;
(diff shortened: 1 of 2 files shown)


src/DataAccess/Services/MeteredPlanSchedulerManagementRepository.cs
```diff
@@ -63,7 +63,7 @@ public int Save(MeteredPlanSchedulerManagement entity)
         if (entity.StartDate.HasValue)
         {
             int minute = entity.StartDate.Value.Minute;
-            if (entity.StartDate.Value.Minute >= 30)
+            if (entity.StartDate.Value.Minute >= 30 || entity.StartDate.Value.ToUniversalTime() < DateTime.UtcNow)
```
msalemcode, line 66: beside adding this logic, can we also modify the NewScheduler.shtml  JavaScript function to also add an hour in future. I notice many ISV click save to current time.

  $(document).ready(function() {
        var d = new Date();
 -->  d.setHours(d.getHours() + 1); <---

jeernest: sounds reasonable, I have added the change

msalemcode (approved):

PR #578, opened 2023-10-18
Fix performance when reading job LastRun

jeernest:
The problem is that the code was repetitively trying to get subscription ID for all scheduled jobs. It means that for many jobs it is called really many times. I have rewritten the SQL so it returns the required information using JOIN, which is much faster.

Also I have implemented a memory cache per subscription ID, because if we have n jobs for every subscription it doesn't make sense to fetch everything n times. The audit log can be large especially if we are talking about daily jobs.


santhoshb-msft:
Hi @jeernest Thank you very much for this contribution, but currently we are not looking to implement caching at this stage as we will be reviewing this in future and improve accordingly. Closing this for now.

PR #581, opened 2023-10-24
Fix last run task

msalemcode:
Change the scheduler task list to return items expected to execute and ignore items already executed.

### DIFF
--- a/src/Services/Services/MeteredPlanSchedulerManagementService.cs
+++ b/src/Services/Services/MeteredPlanSchedulerManagementService.cs
@@ -164,28 +164,27 @@ public class MeteredPlanSchedulerManagementService
     /// Get All Scheduled Metered trigger list
     /// </summary>
     /// <returns>List of Scheduler Manager View</returns>
-    public List<SchedulerManagerViewModel> GetScheduledTasks()
+    public IReadOnlyList<SchedulerManagerViewModel> GetScheduledTasks()
     {
-        List<SchedulerManagerViewModel> schedulerList = new List<SchedulerManagerViewModel>();
-        var allSchedulerViewData = this.schedulerViewRepository.GetAll().OrderBy(s => s.AMPSubscriptionId);
+         List<SchedulerManagerViewModel> schedulerList = new List<SchedulerManagerViewModel>();
+        var allSchedulerViewData = this.schedulerViewRepository.GetAll().OrderBy(s => s.StartDate);
         foreach (var item in allSchedulerViewData)
         {
-            SchedulerManagerViewModel schedulerView = new SchedulerManagerViewModel();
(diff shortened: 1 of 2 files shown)


src/Services/Services/MeteredPlanSchedulerManagementService.cs
```diff
@@ -159,6 +159,43 @@ public List<SchedulerManagerViewModel> GetAllSchedulerManagerList()
...
 
+
+    /// <summary>
+    /// Get All Scheduled Metered trigger list
+    /// </summary>
+    /// <returns>List of Scheduler Manager View</returns>
+    public List<SchedulerManagerViewModel> GetScheduledTasks()
+    {
+        List<SchedulerManagerViewModel> schedulerList = new List<SchedulerManagerViewModel>();
+        var allSchedulerViewData = this.schedulerViewRepository.GetAll().OrderBy(s => s.AMPSubscriptionId);
```
code4clouds, line 170: Wouldn't it be better to sort it with the dates closest to the current time, so they execute closer together and help avoid missing the window?

code4clouds, line 170: this should be a const to avoid alteration and increase performance.

src/Services/Services/MeteredPlanSchedulerManagementService.cs
```diff
@@ -159,6 +159,43 @@ public List<SchedulerManagerViewModel> GetAllSchedulerManagerList()
...
+            schedulerView.SubscriptionName = item.SubscriptionName;
+            schedulerView.AMPSubscriptionId = item.AMPSubscriptionId;
+            schedulerView.Dimension = item.Dimension;
+            schedulerView.Frequency = item.Frequency;
+            schedulerView.Quantity = item.Quantity;
+            schedulerView.StartDate = item.StartDate;
+            schedulerView.NextRunTime = item.NextRunTime;
+            schedulerView.LastRunTime = this.GetSchedulerLastRunTime(item.Id, item.SchedulerName);
+
+            if (!CheckIfSchedulerRun(item.Id, item.SchedulerName))
```
code4clouds, line 187: The IF should be under the foreach then if TRUE assign values then add to the list.  Adding the values first then throwing them away if the condition is not met waste allocation time and impact performance.   

In other words the allocation to the list should be done only after the condition is met.

code4clouds (changes requested):
Code changes need around memory allocaiton and performance.

src/Services/Services/MeteredPlanSchedulerManagementService.cs
msalemcode: Update to sort by start date

code4clouds (approved):

PR #586, opened 2023-10-30
Fetch SMTP password from KeyVault

jeernest:
This change avoids storing the password as plain text in database and instead allows the accelerator to fetch password from Key Vault. Just configure the SMTPPassword setting on UI with the pattern pointing to the keyvault secret:
```
@Microsoft.KeyVault(VaultName=webapp-prefix-kv;SecretName=SMTPPassword)
```


santhoshb-msft:
Hi @jeernest thank you for this contribution, after an internal discussion, I think we want to build something which supports both ways of adding the value plus also there are other scenarios which would break the existing instance functionality and our video guidance should be updated. Due to this, I'm currently closing this, I would like once again thank you for all your contributions, they are definitely valuable.

PR #616, opened 2023-12-05
Send email address from the Email Template

santhoshb-msft:
1. Added to email address from Email Template
2. If the plan level email address is setup then the above to address will be overridden.


neelavarshad:
@santhoshb-msft  Confirming below behavior:
1. Set To Recipients within the Email template in Admin portal to receive emails
2. If recipient email is set null at the plan level, email is automatically sent to the Email recipients specified in the email template.
3. If the recipient email is set at a plan level, then the email is sent to the address at the plan level and the email template recipients are omitted.

neelavarshad (approved):

PR #630, opened 2024-01-05
Fix New Scheduler add hour Javascipt logic

msalemcode:
Current JS logic will pad leading zero then add hour in same time. The logic will break in case of hour=9

The new logic will add the hour first then go through the logic of leading zero


santhoshb-msft (approved):
tested. looks good

Issue #635, opened 2024-01-12
The Deploy.ps1 script gets a IPV6 Public IP instead of and IPV4, causing an error while setting up the SQL firewall rule


Giancarl021:
**Describe the bug**
When running the [`Deploy.ps1`](https://github.com/Azure/Commercial-Marketplace-SaaS-Accelerator/blob/536cbf1d958957d7049bee120bd5a4b00fe8adb8/deployment/Deploy.ps1) file, I get an `FirewallRuleNotIPv4Address`, showing that the service [ifconfig.me](https://ipconfig.me/ip) is returning the wrong IP type.

**To Reproduce**
1. Execute the [`Deploy.ps1`](https://github.com/Azure/Commercial-Marketplace-SaaS-Accelerator/blob/536cbf1d958957d7049bee120bd5a4b00fe8adb8/deployment/Deploy.ps1) PowerShell file;
2. On the [line 350](https://github.com/Azure/Commercial-Marketplace-SaaS-Accelerator/blob/536cbf1d958957d7049bee120bd5a4b00fe8adb8/deployment/Deploy.ps1#L350C2-L350C70), it will grab the IPV6 IP address, instead of the expected IPV4;
3. An `FirewallRuleNotIPv4Address` error will be thrown

**Expected behavior**
The correct IP type to be returned by the [ifconfig.me](https://ipconfig.me/ip) service, which will lead to a clean execution.

**Screenshots**
```
      ➡️ Add SQL Server Firewall rules
{
  "endIpAddress": "0.0.0.0",
  "id": "/subscriptions/***/resourceGroups/SaaS-Accelerator/providers/Microsoft.Sql/servers/***/firewallRules/AllowAzureIP",
  "name": "AllowAzureIP",
  "resourceGroup": "***",
  "startIpAddress": "0.0.0.0",
  "type": "Microsoft.Sql/servers/firewallRules"
}
      ➡️ Running in local environment - Add current IP to firewall
(FirewallRuleNotIPv4Address) '0000:0000:0000:0000:0000:0000:0000:0000' is not a valid IPv4 address.
Code: FirewallRuleNotIPv4Address
Message: '0000:0000:0000:0000:0000:0000:0000:0000' is not a valid IPv4 address.
```

**Environment (please complete the following information):**
 - Are you using the CloudShell? No
 - If not using the CloudShell:
    - OS: Ubuntu 22.04.3 LTS
    - PowerShell version: 7.4.0
[...]

PR #636, opened 2024-01-12
Change service that gives the public IP of the client

Giancarl021:
Related to #635.

This Pull Request changes the service that gives the client it's public IP address.
The change is made to avoid an IPV6 to be fed into the firewall rule at the deployment of the SQL Server.


santhoshb-msft:
@Giancarl021 Can I please confirm, this issue you are fixing during installation from your local env right?

Giancarl021:
> @Giancarl021 Can I please confirm, this issue you are fixing during installation from your local env right?

Yes. I understand that the issue might be caused by my local environment. The fix is more of a suggestion if considered interesting.

santhoshb-msft:
Thank you @Giancarl021 , I will let @msalemcode review this.

msalemcode (approved):

PR #638, opened 2024-01-16
Add dockerfile 

msalemcode:
This PR will add dockerfile for admin and customer sites.

In order to run the container properly, startup.cs files were update to enforce https redirect.


santhoshb-msft:
We had a brief discussion and will continue further

PR #650, opened 2024-02-02
Added logic for missing terms dates during activation and renewal.

santhoshb-msft:
Added logic for missing terms dates during activation and renewal.


santhoshb-msft:
putting on hold for the moment.

PR #662, opened 2024-02-23
set the cookie expiration time

santhoshb-msft:


neelavarshad (approved):
Approved cookie expiry

neelavarshad:
Approving:
1. Cookie expiry time is set to 60 minutes with sliding expiration
2. Updated startup.cs file with validated antiforgerytoken
3. Ignore antiforgery token for webhook
4. Referencing antiforgery token from form element

PR #667, opened 2024-02-29
added validations to only allowed users to fetch subscriptions

santhoshb-msft:


neelavarshad (approved):
Approving change for validation of user's subscription

PR #670, opened 2024-03-12
Allow Existing KeyVault to process withput breaking

msalemcode:
If user passed KeyVault name or Accelerator deployed over existing resouces, Script will exist due to existing KeyVault and direct user to purge KeyVault.

**Solution:**
1- If user provided KeyVault name, Deployment script will by pass the validation. It is user responsibility to provide existing valid KeyVault name
2- If Deployment used auto generated keyvault name then there is two different actions:
- If KeyVault exists then it will pass and proceed with deployment. 
- If KeyVault does not exist then it will check if the KeyVault was deleted before or not.

### DIFF
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -54,7 +54,7 @@ if($KeyVault -eq "")
    $KeyVault=$WebAppNamePrefix+"-kv"
 
    # Check if the KeyVault exists under resource group
-   $kv_check=$(az keyvault show -n $KeyVault -g $ResourceGroupForDeployment)     
+   $kv_check=$(az keyvault show -n $KeyVault -g $ResourceGroupForDeployment) 2>$null    
 
    # If KeyVault does not exist under resource group, then we need to check if it deleted KeyVault
    if($kv_check -eq $null)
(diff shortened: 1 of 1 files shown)


deployment/Deploy.ps1
```diff
@@ -48,7 +48,38 @@ if ($SQLDatabaseName -eq "") {
 
 if($KeyVault -eq "")
```
neelavarshad, line 49: validation will not run if the user provides a keyvault that already exists, correct?

neelavarshad:
@msalemcode The validation is right, although the script could still fail, if the user provides the name of a keyvault that already exists. 
At the time of deploying azure resource (after preparing publish files), we are calling az create again on the same keyvault name, there should be a check to ensure we are not recreating keyvault if it already exists.

deployment/Deploy.ps1
msalemcode: Correct, the idea if the user provides a keyvault then we will skip over the validation. Validation will be good only in case of keyvault is soft deleted.

neelavarshad: Thanks for the clarification. In this case, publisher is responsible for ensuring that the keyvault is not soft deleted.

neelavarshad (approved):

PR #671, opened 2024-03-12
Add logic to validate token on webhook notification

santhoshb-msft:
Add logic to validate token on webhook notification


src/DataAccess/Migrations/20240312055030_baseline751.cs
neelavarshad, line 1: @santhoshb-msft  I'll reach out separately for the DB functions.

santhoshb-msft: Discussed internally

neelavarshad (approved):

PR #674, opened 2024-03-15
Update app log message for notification URL

santhoshb-msft:
Update app log message for notification URL

### DIFF
--- a/src/AdminSite/Controllers/ApplicationConfigController.cs
+++ b/src/AdminSite/Controllers/ApplicationConfigController.cs
@@ -2,16 +2,13 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text.Json;
-using System.Web;
 using Marketplace.SaaS.Accelerator.DataAccess.Contracts;
 using Marketplace.SaaS.Accelerator.DataAccess.Entities;
-using Marketplace.SaaS.Accelerator.DataAccess.Services;
+using Marketplace.SaaS.Accelerator.Services.Helpers;
 using Marketplace.SaaS.Accelerator.Services.Services;
 using Marketplace.SaaS.Accelerator.Services.Utilities;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.Logging;
 
 namespace Marketplace.SaaS.Accelerator.AdminSite.Controllers;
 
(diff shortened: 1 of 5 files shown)


src/Services/Helpers/UrlValidator.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Marketplace.SaaS.Accelerator.Services.Helpers;
+
+/// <summary>
+/// URL Validator.
+/// </summary>
+public class UrlValidator
+{
+    /// <summary>
+    /// Validates the URLfor HTTPS.
```
007zszmz, line 11: missing a white space.

src/Services/Helpers/UrlValidator.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Helps validate if the URL is a valid HTTPS URL.
+    /// </summary>
+    public static bool IsValidUrlHttps(string url)
+    {
+        return Uri.TryCreate(url, UriKind.Absolute, out Uri uriResult) && 
+               uriResult.Scheme == Uri.UriSchemeHttps && 
+               uriResult.Port == 443;
```
007zszmz, line 18: how about if we test locally with a different port number than 443?

santhoshb-msft: @007zszmz I think the idea is we dont them to use anything other than https. we can disable this while testing if we need to. Happy to chat internally

msalsouri (commented):
review

007zszmz (approved):

src/AdminSite/Controllers/ApplicationConfigController.cs
```diff
@@ -119,6 +116,13 @@ public IActionResult ApplicationConfigDetails(int Id)
     public IActionResult ApplicationConfigDetails(ApplicationConfiguration appConfig)
     {
         appConfig.Value = appConfig.Value ?? string.Empty;
+        
+        //check with the config is webhnotifcation url then validate if its proper url
+        if (appConfig.Name == "WebNotificationUrl" && !UrlValidator.IsValidUrlHttps(appConfig.Value))
```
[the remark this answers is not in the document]
santhoshb-msft: updated accordingly.

src/Services/Helpers/UrlValidator.cs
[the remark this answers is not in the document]
santhoshb-msft: fixed the same

PR #676, opened 2024-03-18
Upgraded Chrome driver from 120 to 122

santhoshb-msft:
Upgraded Chrome driver from 120 to 122


santhoshb-msft:
skipping the approval due to the change is only relevant to UI test project.

PR #678, opened 2024-03-21
added vnet integrations

santhoshb-msft:
1. Add Vnet and Subnets
2. Added delegations and service endpoints
3. Locked KV and SQL to Vnet


neelavarshad (approved):
Approving VNET integrations to keyvault

PR #680, opened 2024-03-28
credencials have been added to Admin & Customer sites and renamed the…

msalsouri:
credentials have been added to Admin & Customer sites and renamed the appsettings.Development.json


bkhabazan:
Changes need to be added to a forked library, and should not be pushed to the open source version

PR #688, opened 2024-04-09
Updated to .net7 to run on azure

WeAreJigital:


alexdrenea:
.net 7 is STS.

I would wait for .net8

PR #711, opened 2024-06-01
Move modals outside of form elements

michaellperry:
The modals in both the admin and landing page subscription lists were not working. The spinner would keep spinning indefinitely after the user pressed the button.

The cause was that the form element that the JavaScript attempted to submit did not exist. The form was nested within another form. As a result, the browser removed it from the DOM.

This change moves the modal container outside of the outer form element so that it's inner form would not be nested.

### DIFF
--- a/src/AdminSite/Views/Home/Subscriptions.cshtml
+++ b/src/AdminSite/Views/Home/Subscriptions.cshtml
@@ -128,25 +128,25 @@
                     </div>
                 </div>
             </div>
-            <!-- Modal -->
-            <div class="modal" id="myModal" tabindex="-1">
-                <div class="modal-dialog">
-                    @* <div class="modal-content">
-                        <div class="modal-header">
-                            <h5 class="modal-title" id="exampleModalLabel">Modal title</h5>
-                            <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
-                        </div> *@
-                        @* <div class="modal-body">
-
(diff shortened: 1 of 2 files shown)


michaellperry:
@microsoft-github-policy-service agree company="Jinaga LLC"

santhoshb-msft:
@neelavarshad please review as needed.

santhoshb-msft:
Hi @michaellperry can you provide with some repro steps please, im trying but unable

santhoshb-msft:
reaching out again to see any repro steps for the issue please.

michaellperry:
Administer your SaaS offer in the SaaS Accelerator admin portal.
image

Go to the Application Config section. Select the "AcceptSubscriptionUpdates" setting.
image

Edit the setting and change the value to "true".
image

Go to the "Plans" page. Ensure that you have more than one plan.
image

Next log into the Azure Portal. Subscribe to your offer.
image

Activate the subscription and go to the subscription list.
image

Select "Change Plan" in the menu next to the subscription.
image

The change plan dialog appears. Select a different plan.
image

Click the "Change Plan" button. It does not respond.
image

fthorntonai:
WIP-9/30/2024

fthorntonai (approved):
This solution is acceptable but PR should've been submitted using request format. We should also review the design and flow of the dialog for improvements.

santhoshb-msft:
will merge after release 8

PR #727, opened 2024-07-15
Ability to have emails flowing without needing to set at individual plan level

santhoshb-msft:
1. Removed condition to setup email configuration at plan level instead moved to email template levelas it would be more global and work for all plans.
2. Added email send/error logs to application logs.
3. Added try/catch/using for email send.
4. Updated read me instructions.


neelavarshad:
Adding @bkhabazan  and @007zszmz as optional reviewers

src/Services/Services/SMTPEmailService.cs
```diff
@@ -32,36 +47,55 @@ public SMTPEmailService(IApplicationConfigRepository applicationConfigRepository
...
+                    //send message
+                    smtp.Send(mail);
+                    this.applicationLogService.AddApplicationLog($"{emailContent?.Subject}: Email sent succesfully!").ConfigureAwait(false);
+                }
+            }
+            catch (SmtpException smtpEx)
             {
-                mail.To.Add(new MailAddress(multimailid));
+                // Handle SMTP specific exceptions here
+                applicationLogService.AddApplicationLog($"SMTP Error: {emailContent?.Subject} {smtpEx.Message}").ConfigureAwait(false);
```
007zszmz, line 92: not awaited

007zszmz (approved):

src/Services/Services/SMTPEmailService.cs
santhoshb-msft: agree @007zszmz but to await this I might need to make changes up the chain which is going to be a lot bigger change, and these calls are pretty much fire-n-forget scenario, hence not too worried about. What do you think?

src/AdminSite/Views/Plans/Index.cshtml
neelavarshad, line 1: Is the Plan ID just a good to have addition?

santhoshb-msft: I think so, @neelavarshad , just sometimes with same name it can be confusing and with privateoffer v2 auto planid set it just helps, pic for reference
image

src/Services/Services/SMTPEmailService.cs
007zszmz: i agree. just want to call it out, so this is a deliberate decision. worse case scenario is we lost error logs for troubleshooting.

santhoshb-msft: yes that right, thanks @007zszmz ! Resolving this for now.

cc: @neelavarshad

neelavarshad (approved):

PR #756, opened 2024-08-26
Add file name and executable file validation.

bkhabazan:
Added a check to ensure that the file name does not contain more than one period. If it does, an error message is set in `TempData` and the user is redirected to the "Index" action.

Introduced a new block of code that reads the first two bytes of the uploaded file to determine if it is an executable file (by checking for the "MZ" header). If the file is identified as an executable, an error message is set in `TempData` and the user is redirected to the "Index" action.

### DIFF
--- a/src/AdminSite/Controllers/ApplicationConfigController.cs
+++ b/src/AdminSite/Controllers/ApplicationConfigController.cs
@@ -171,6 +171,25 @@ public class ApplicationConfigController : BaseController
             }
 
             var fileExtension = Path.GetExtension(file.FileName).ToLowerInvariant();
+            int dotCount = Path.GetFileNameWithoutExtension(file.FileName).Count(f => f == '.');
+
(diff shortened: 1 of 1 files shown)


v-adeadepoju (changes requested):
https://aka.ms/dotnet/sdk-not-found
Exception: /home/adedayo/Commercial-Marketplace-SaaS-Accelerator/deployment/Deploy.ps1:112
Line |
 112 |      Throw "🛑 Dotnet 6 not installed. Install dotnet6 and re-run the  …
     |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | 🛑 Dotnet 6 not installed. Install dotnet6 and re-run the script.

santhoshb-msft:
lets work on this PR once 8 is out

PR #778, opened 2024-11-06
Change String-Between

ty05:
The ISV encountered an error while attempting to upgrade the SaaSAccelerater. The error occurs in the Upgrade.ps1 script, which tries to parse User and Pass from the KeyVault information. However, the KeyVault only contains AAD authentication details and does not include User or Pass.

I changed the secret in the Key Vault to the SQL database connection string. However, the String-Between function returns the server name under "User", which causes the login error.

I modified the String-Between function, and it works, at least in my case.


ty05:
@microsoft-github-policy-service agree

PR #812, opened 2025-04-07
Fix KV Private Endpoint DNS Name

alexdrenea:
Fixing a small issue with Private Endpoint for KeyVault so that traffic actually flows though the private network between webapp and KV


CB-Vista:
This drove me nuts. When you deploy the accelerator it all works for a while, then stops after some period of time. Took me hours to figure out it was using the wrong private DNS name
https://learn.microsoft.com/en-us/azure/private-link/private-endpoint-dns#security

PR #847, opened 2025-07-02
fix vulnerable dependencies

Retrieval6945:


Retrieval6945:
@microsoft-github-policy-service agree

jd1100 (commented):
Updating packages to fix known cves

